PR #1, opened 2021-08-17
Made `Setup` optional

djspiewak:


lambda/shared/src/main/scala/feral/lambda/IOLambda.scala
```diff
@@ -24,14 +24,15 @@ import cats.syntax.all._
...
+abstract class IOLambda[Event, Result](
     implicit private[lambda] val decoder: Decoder[Event],
     private[lambda] val encoder: Encoder[Result]
-) extends IOLambdaPlatform[Setup, Event, Result] {
+) extends IOLambdaPlatform[Event, Result] {
 
   protected def runtime: IORuntime = IORuntime.global
 
-  def setup: Resource[IO, Setup]
+  protected type Setup
```
armanbilge, line 34: Cool, I didn't realize a concrete class/object can just decline to define this and the compiler doesn't complain.

armanbilge (approved):
Btw I still feel weird making this a `Resource` since we are violating an implied promise to run the finalizer. But I see the practicality of it.

Also feels weird to be on this side of the table :P

PR #3, opened 2021-08-20
Make Lambda Context an http4s attribute

armanbilge:
Uses an ordinary `HttpRoutes` instead of `ContextRoutes` and provides the lambda `Context` as an attribute in the `Request`.


armanbilge:
While we're at it, let's include the trigger `ApiGatewayProxyEventV2` as an attribute too. So now a user can still access everything should they need to.

PR #4, opened 2021-08-20
Create Cloudflare Worker module

armanbilge:
Still WIP but I feel quite excited about this one to try implementing the [serverless concept](https://gist.github.com/djspiewak/37a4ea0d7a5237144ec8b56a76ed080d). Cloudflare Workers are simple and very approachable IMO, come out-of-the-box with a key-value "database" API, and can be deployed via a single API call. So I think this is perfect for implementing a quick serverless app :)


build.sbt
```diff
@@ -39,15 +39,26 @@ lazy val root =
       lambdaEvents.js,
       lambdaEvents.jvm,
       lambdaApiGatewayProxyHttp4s.js,
-      lambdaApiGatewayProxyHttp4s.jvm)
+      lambdaApiGatewayProxyHttp4s.jvm,
+      cloudflareWorker
```
bpholt, line 43: Is there any reason not to add `core.jvm` and `core.js` to this list? It makes testing things locally a little easier and we may need to anyway to make sure those artifacts get published.

armanbilge: Woops, just forgot to! Good catch :)

armanbilge:
@darkfrog26 expressed an interest in this so I'm going to try and pick it up again. It's due for a bit of a re-design after all the learnings from the MTL-lambda.

PR #5, opened 2021-08-23
Create core module

armanbilge:
This is cherry-picked out of #4. Creates a core module containing `IOSetup` to abstract the one-time-resource-acquisition logic.

### DIFF
--- a/core/src/main/scala/feral/IOSetup.scala
+++ b/core/src/main/scala/feral/IOSetup.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


core/src/main/scala/feral/IOSetup.scala
```diff
@@ -0,0 +1,45 @@
...
+
+package feral
+
+import cats.effect.unsafe.IORuntime
+import cats.effect.kernel.Resource
+import cats.effect.IO
+import cats.effect.kernel.Deferred
+import cats.syntax.all._
+
+private[feral] trait IOSetup {
```
armanbilge, line 25: Do we want to expose this outside of `feral`?

core/src/main/scala/feral/IOSetup.scala
```diff
@@ -0,0 +1,45 @@
...
+
+  protected def runtime: IORuntime = IORuntime.global
+
+  protected type Setup
+  protected def setup: Resource[IO, Setup] = Resource.pure(null.asInstanceOf[Setup])
+
+  private[feral] final lazy val setupMemo: IO[Setup] = {
```
armanbilge, line 32: @bpholt and I changed this to `def` / `lazy val` to avoid initialization errors, undoing your change In #1 to make `setup` a `val`. Is there a better way to handle this?

bpholt (approved):

PR #6, opened 2021-09-01
add CloudFormation Custom Resource

bpholt:
I updated one of our CloudFormation custom resources to use this draft implementation: https://github.com/Dwolla/postgresql-init-custom-resource/pull/6


build.sbt
```diff
@@ -97,3 +97,16 @@ lazy val lambdaApiGatewayProxyHttp4s = crossProject(JSPlatform, JVMPlatform)
     )
   )
   .dependsOn(lambda, lambdaEvents)
+
+lazy val cloudformationCustomResource = crossProject(JSPlatform, JVMPlatform)
+  .crossType(CrossType.Pure)
+  .in(file("cloudformation-custom-resource"))
+  .settings(
+    name := "feral-cloudformation-custom-resource",
```
armanbilge, line 105: Obviously still figuring this stuff out, but so far I'm namespacing AWS stuff as `feral-lambda`. (This is a lambda, right? Or should the namespace we use be aws?)

cloudformation-custom-resource/src/main/scala/feral/cloudformation/package.scala
```diff
@@ -0,0 +1,111 @@
+package feral
+
+import cats.implicits._
+import shapeless.tag.@@
+import io.circe._
+import io.circe.syntax._
+import io.circe.generic.semiauto._
+
+package object cloudformation {
```
armanbilge, line 9: Can some of this stuff be moved to the `events` module maybe? Perhaps a bit over-modulization, but I think its valuable to offer our case class encodings of the basic AWS event types so people can use them with their own custom `IOLambda`s without buying into our implementations.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,91 @@
...
+  extends IOLambda[CloudFormationCustomResourceRequest[Input], Unit]
+    with Http4sClientDsl[IO] {
+  override protected type Setup = (Client[IO], CfnSetup)
+
+  def cfnSetup: Resource[IO, CfnSetup]
+  def handler(setup: CfnSetup): CloudFormationCustomResource[IO, Input, Output]
+
+  override protected def setup: Resource[IO, Setup] =
+    (EmberClientBuilder.default[IO].build, cfnSetup).parTupled
+
```
armanbilge, line 35: I think we should make `setup` final and delegate `client` to an overridable lazy val that defaults to ember.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,91 @@
...
+import java.io.{PrintWriter, StringWriter}
+
+trait CloudFormationCustomResource[F[_], Input, Output] {
+  def createResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+  def updateResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+  def deleteResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+}
+
+abstract class CloudFormationCustomResourceHandler[CfnSetup, Input : Decoder, Output: Encoder]
```
armanbilge, line 25: @djspiewak likes to make this `CfnSetup`-like stuff into type members, although I'm not sure if that's appropriate here.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,91 @@
...
+  extends IOLambda[CloudFormationCustomResourceRequest[Input], Unit]
+    with Http4sClientDsl[IO] {
+  override protected type Setup = (Client[IO], CfnSetup)
+
+  def cfnSetup: Resource[IO, CfnSetup]
+  def handler(setup: CfnSetup): CloudFormationCustomResource[IO, Input, Output]
```
armanbilge, line 31: I'm a little confused why we need these seperate. Is it possible to just ask for a `Resource[CloudFormationCustomResource[IO, Input, Output]`? Like I did in the http4s lambda, asking for a `Resource[HttpRoutes[IO]]` I think.

armanbilge (commented):
Thanks!! Really excited about this. Still need to take a closer look but just added some quick thoughts.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/package.scala
bpholt: I went back and forth on using the `events` module. I see where you're coming from by allowing others to use the models, but it also means as a user of just the CloudFormation Custom Resource class, I have a lot of models in scope that actually aren't relevant to that task. Tradeoffs either way, I think.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/CloudFormationCustomResource.scala
bpholt: Can you elaborate on this? I'm not entirely following how that fits together.

cloudformation-custom-resource/src/main/scala/feral/cloudformation/CloudFormationCustomResource.scala
bpholt: How does the implementation of `Resource[CloudFormationCustomResource[IO, Input, Output]` get access to the materialized `CfnSetup`?

armanbilge:
@bpholt Apologies I wasn't so clear in my comments above, I think I'm imagining an API like this without `CfnSetup`, hopefully this makes a bit more sense?

```scala
abstract class CloudFormationCustomResourceHandler[Input: Decoder, Output: Encoder]
  extends IOLambda[CloudFormationCustomResourceRequest[Input], Unit] {
  
  // Users can override if they wish
  def client: Resource[IO, Client[IO]] = EmberClientBuilder.default[IO].build
  
  // client is the only shared setup, users can setup anything else they want in this resource
  def handler(client: Client[IO]): Resource[CloudFormationCustomResource[IO, Input, Output]
  
  // Now we combine the above to get everything we need in Setup
  type Setup = (Client[IO], CloudFormationCustomResource[IO, Input, Output])
  override final def setup: Resource[IO, Setup] = client.flatMap(client => (client, handler(client)))

  // ...
}
```

cloudformation-custom-resource/src/main/scala/feral/cloudformation/package.scala
armanbilge: I agree, certainly tradeoffs either way, definitely open to ideas for how to best tackle this. FWIW the `events` module is akin to [aws-lambda-java-events](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events) and [aws-lambda/trigger](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/aws-lambda/trigger). Maybe we could scope this a bit better with additional packages?

Also, to explain my thinking a bit, I tend to determine modules by their dependencies. Taking the JS implementations as an example:
- `core`: CE3
- `lambda`: `core` + `circe`
- `events`: `circe-generic` (i.e. shapeless :)
- `lambdaApiGatewayProxyHttp4s`: `lambda` + `events` + `http4s-core`
- `lambdaCloudFormationCustomResource`: `lambda` + `events` + `http4s-ember-client`

So, the `events` module essentially lets users write lambdas without having http4s (or even CE3!) in scope if that's not relevant to what they want to do either.

I think it ultimately comes down to priorities, if we pretty much assume people will be doing things our way then I totally agree with you, no need for an `events` module and we can keep things scoped.

bpholt: My preference is to leave these things in the CloudFormation module, so I left this alone for now. I think if people aren't doing things our way they're more likely to just use aws-lambda-java-events or whatever, and if they are using our core then having a more focused set of model classes makes sense. But either way is fine, so if there's consensus to move them into events, I can do that too.

build.sbt
```diff
@@ -97,3 +99,21 @@ lazy val lambdaApiGatewayProxyHttp4s = crossProject(JSPlatform, JVMPlatform)
     )
   )
   .dependsOn(lambda, lambdaEvents)
+
+lazy val cloudformationCustomResource = crossProject(JSPlatform, JVMPlatform)
```
armanbilge, line 103:
```suggestion
lazy val lambdaCloudformationCustomResource = crossProject(JSPlatform, JVMPlatform)
```

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,105 @@
...
+
+  def handler(client: Client[IO]): Resource[IO, CloudFormationCustomResource[IO, Input, Output]]
+
+  override def apply(event: CloudFormationCustomResourceRequest[Input],
+                     context: Context,
+                     setup: Setup): IO[Option[Unit]] =
+    (event.RequestType match {
+      case CreateRequest => setup._2.createResource(event)
+      case UpdateRequest => setup._2.updateResource(event)
+      case DeleteRequest => setup._2.deleteResource(event)
```
armanbilge, line 60: Should we pass the `Context` object to the handler as well? I think you said you've never used it, but maybe there's something useful in there e.g. `awsRequestId` that someone might want to use for logging?

armanbilge (approved):
Thanks for finishing this up! Just some nits, but nothing to block on :)

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,105 @@
...
+  def deleteResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+}
+
+abstract class CloudFormationCustomResourceHandler[Input : Decoder, Output: Encoder]
+  extends IOLambda[CloudFormationCustomResourceRequest[Input], Unit]
+    with Http4sClientDsl[IO] {
+  type Setup = (Client[IO], CloudFormationCustomResource[IO, Input, Output])
+
+  override final def setup: Resource[IO, Setup] =
+    client.mproduct(handler)
```
armanbilge, line 47: TIL `mproduct` :)

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -0,0 +1,105 @@
...
+
+trait CloudFormationCustomResource[F[_], Input, Output] {
+  def createResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+  def updateResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+  def deleteResource(event: CloudFormationCustomResourceRequest[Input]): F[HandlerResponse[Output]]
+}
+
+abstract class CloudFormationCustomResourceHandler[Input : Decoder, Output: Encoder]
+  extends IOLambda[CloudFormationCustomResourceRequest[Input], Unit]
+    with Http4sClientDsl[IO] {
```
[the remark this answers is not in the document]
bpholt: Good idea. I almost always use `Http4sClientDsl[F]` so I kind of forgot the import approach was available 🙂

bpholt:
Yes please, a snapshot would be great

armanbilge:
Sorry, just saw this. Do you have keys you can install that can publish under `org.typelevel`? Otherwise I can publish under my own group.

PR #33, opened 2021-10-20
Make `Lambda` generic in effect type

armanbilge:
Maybe just me, but building all this stuff in terms of straight `IO` was bugging me. This takes a step away from this to make the effect type generic.

I started by defining a `trait Feral[F[_]]` which encapsulates:
* an `Async[F]`
* a memoized `F[Setup]` for a `Resource[F, Setup]`
* a `Dispatcher[F]`

The idea here is not so much to provide a "real" `Dispatcher` (i.e. that supervises its fibers) but as a convenience abstraction for the capability to unsafely run effects. Technically this is an abuse, but here we are setting up a `Resource` that we have no intention of closing, so yolo.

Thus we can provide a convenience `IOFeral` that implements the `Dispatcher[IO]` directly in terms of `IO`'s unsafe methods and the `IORuntime`.

From here it is straightforward to make everything else generic in `F`. `IOLambda` becomes just `Lambda`, and `IOLambda` is provided as a convenience mix-in.

Looking forward to thoughts on this.


bpholt:
On its face, I think this is a good idea. I almost always write code in an abstract `F[_]` so working in concrete `IO` for the CloudFormation custom resource felt weird to me too.

I pulled this in locally and rebased against `main` so I could try it with the CFN custom resource. Without actually deploying it, it seems to work fine, especially after I added a 

```scala
abstract class IOCloudFormationCustomResourceHandler[Input : Decoder, Output: Encoder]
  extends CloudFormationCustomResourceHandler[IO, Input, Output]
    with IOFeral
```

to make it a little more ergonomic when implementing the actual Lambda class.

armanbilge:
Yes, I'm all for defining those convenience versions 👍 btw feel free to push your branch here, save me the rebase 😉

armanbilge:
Newer ideas in https://github.com/typelevel/feral/pull/45.

PR #44, opened 2021-11-08
add tracing support with Natchez

bpholt:
I've been using this with a locally published version of https://github.com/christiankjaer/natchez/pull/1 and https://github.com/tpolecat/natchez/pull/427 and it seems to work.

My thought is that we can bake in no-op tracing, and then the user can mix in a trait that implements tracing for their preferred backend. It doesn't compile here yet (because of the missing artifacts) but the `XRayTracing` trait is an example of how that could work. WDYT?

### DIFF
--- a/lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
+++ b/lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
@@ -22,6 +22,7 @@ import cats.effect.SyncIO
 import feral.lambda.events.ApiGatewayProxyEventV2
 import feral.lambda.events.ApiGatewayProxyStructuredResultV2
 import fs2.Stream
+import natchez.Trace
 import org.http4s.Charset
 import org.http4s.Header
 import org.http4s.Headers
@@ -47,7 +48,8 @@ abstract class ApiGatewayProxyHttp4sLambda
   override final def apply(
       event: ApiGatewayProxyEventV2,
       context: Context,
-      routes: HttpRoutes[IO]): IO[Some[ApiGatewayProxyStructuredResultV2]] =
+      routes: HttpRoutes[IO])(
+      implicit T: Trace[IO]): IO[Some[ApiGatewayProxyStructuredResultV2]] =
     for {
       method <- IO.fromEither(Method.fromString(event.requestContext.http.method))
       uri <- IO.fromEither(Uri.fromString(event.rawPath))
--- a/lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
+++ b/lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
@@ -16,23 +16,13 @@
 
 package feral.lambda
 
-import cats.effect.IO
-import cats.effect.Resource
-import cats.effect.SyncIO
-import feral.lambda.events.ApiGatewayProxyEventV2
-import feral.lambda.events.ApiGatewayProxyStructuredResultV2
+import cats.effect.{IO, Resource, SyncIO}
+import feral.lambda.events.{ApiGatewayProxyEventV2, ApiGatewayProxyStructuredResultV2}
 import fs2.Stream
-import natchez.Trace
-import org.http4s.Charset
-import org.http4s.Header
-import org.http4s.Headers
-import org.http4s.HttpRoutes
-import org.http4s.Method
-import org.http4s.Request
-import org.http4s.Response
-import org.http4s.Uri
-import org.typelevel.vault.Key
-import org.typelevel.vault.Vault
+import natchez.http4s.NatchezMiddleware
+import natchez.{Span, Trace}
+import org.http4s.{Charset, Header, Headers, HttpRoutes, Method, Request, Response, Uri}
+import org.typelevel.vault.{Key, Vault}
 
 abstract class ApiGatewayProxyHttp4sLambda
     extends IOLambda[ApiGatewayProxyEventV2, ApiGatewayProxyStructuredResultV2] {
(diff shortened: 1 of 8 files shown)


lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
```diff
@@ -47,7 +48,8 @@ abstract class ApiGatewayProxyHttp4sLambda
   override final def apply(
       event: ApiGatewayProxyEventV2,
       context: Context,
-      routes: HttpRoutes[IO]): IO[Some[ApiGatewayProxyStructuredResultV2]] =
+      routes: HttpRoutes[IO])(
+      implicit T: Trace[IO]): IO[Some[ApiGatewayProxyStructuredResultV2]] =
```
armanbilge, line 52: How does this `Trace` propagate to userland code? Apologies if I'm missing something.

armanbilge (commented):
This is cool! Still gotta wrap my head around all the pieces 😅

lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
bpholt: No, good catch. I've only made changes to `IOLambda` and the `IOLambdaPlatform`s to make sure it worked at that level. We'd need to add the implicit `Trace` to the other abstract methods if we want to do this consistently.

armanbilge: Aha, I thought it might be using some `IOLocal` magic :)

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -59,7 +59,7 @@ abstract class CloudFormationCustomResourceHandler[Input: Decoder, Output: Encod
   override def apply(
       event: CloudFormationCustomResourceRequest[Input],
       context: Context,
-      setup: Setup): IO[Option[Unit]] =
+      setup: Setup)(implicit T: Trace[IO]): IO[Option[Unit]] =
```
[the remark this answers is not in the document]
bpholt: I added `(implicit T: Trace[IO])` to the methods on `CloudFormationCustomResource` above.

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -26,11 +28,21 @@ import scala.scalajs.js.|
...
+    Resource
+      .pure(Context.fromJS(context))
+      .flatMap { (context: Context) =>
+        traceRootSpan(context.functionName).evalMap(ioTrace).evalMap { implicit trace =>
+          for {
+            setup <- setupMemo
+            event <- IO.fromEither(decodeJs[Event](event))
+            result <- apply(event, context, setup)
+          } yield result.map(_.asJsAny).orUndefined
+        }
```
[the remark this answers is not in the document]
bpholt: I tweaked this section

lambda-api-gateway-proxy-http4s/src/main/scala/feral/lambda/ApiGatewayProxyHttp4sLambda.scala
bpholt: I need to think a little more on how the `ApiGatewayProxyHttp4sLambda` fits in here. (Open to suggestions if you see a way to do it with the current structure.)

We may need to change `routes` to be `def routes: Resource[IO, HttpRoutes[Kleisli[IO, Span[IO], *]]]`? But even that doesn't fit perfectly if all we have here is the implicit `Trace[IO]`. Not sure yet.

armanbilge: There's this https://github.com/tpolecat/natchez-http4s

bpholt: I kind of punted and added the Natchez server middleware to `ApiGatewayProxyHttp4sLambda`. The main problem I see is that doesn't make the `Trace[IO]` capability available to anything defined in `def routes: Resource[IO, HttpRoutes[IO]]`, meaning I'm not sure how application code could open its own spans or propagate the trace onwards if it makes calls to other services.

@armanbilge it looked like you started a new branch that (from the commit messages) might be reworking the structure a little bit to help with this. I'm going to pause here to see what you come up with. 🙂

lambda/shared/src/main/scala/feral/lambda/XRayTracing.scala
```diff
@@ -0,0 +1,43 @@
...
+  private def traceEntryPoint: Resource[IO, EntryPoint[IO]] =
+    Resource.eval(Random.scalaUtilRandom[IO]).flatMap { implicit random =>
+      Resource.eval(XRayEnvironment[IO].daemonAddress).flatMap {
+        case Some(addr) => XRay.entryPoint[IO](addr)
+        case None => XRay.entryPoint[IO]()
+      }
+  }
+
+  override protected def traceRootSpan(name: String): Resource[IO, Span[IO]] =
+    Resource.eval(XRayEnvironment[IO].kernelFromEnvironment).flatMap { kernel =>
```
bpholt, line 40: This strategy for extracting the kernel might only apply to normal AWS Lambdas—e.g. for `ApiGatewayProxyHttp4sLambda` the kernel should probably be extracted from the incoming HTTP request headers, not the Lambda's environment. 🤔

bpholt:
I'm going to close this PR in favor of one of the other open tracing PRs—it has served its purpose 🙂

PR #50, opened 2021-11-17
Tracing doodles

armanbilge:
Fork of https://github.com/typelevel/feral/pull/45 for me to doodle with how to integrate natchez with feral.


lambda-natchez/src/main/scala/feral/lambda/natchez/TracedLambda.scala
```diff
@@ -0,0 +1,33 @@
...
+package feral.lambda.natchez
+
+import cats.effect.kernel.MonadCancelThrow
+import feral.lambda.Lambda
+import natchez.EntryPoint
+import natchez.Trace
+
+object TracedLambda {
+  def apply[F[_]: MonadCancelThrow, G[_], Event: HasKernel, Result](entryPoint: EntryPoint[F])(
+      lambda: Trace[G] => Lambda[G, Event, Result])(
```
armanbilge, line 26: I'm not sure yet how I feel about injecting `Trace[G]` like this. It's a bit awkward, but I think it's a cleaner way to support `IOLocal`-based tracing, or tracing in a general `G`.

lambda-natchez/src/main/scala/feral/lambda/natchez/HasKernel.scala
```diff
@@ -0,0 +1,35 @@
...
+trait HasKernel[Event] {
+  def extract(event: Event): Kernel
+}
+
+object HasKernel extends HasKernelLowPriority {
+  @inline def apply[E](implicit hk: HasKernel[E]): hk.type = hk
+
+  implicit val apiGateProxyEventV2Kernel: HasKernel[ApiGatewayProxyEventV2] =
+    e => Kernel(e.headers)
+}
```
armanbilge, line 31: One of the nice things about having all the events together in the same module is it makes it easy to provide these implicits. Although I'm doubtful if any will have a special implementation except this one.

lambda-natchez/src/main/scala/feral/lambda/natchez/TracedLambda.scala
```diff
@@ -16,18 +16,26 @@
...
+      lambda: Trace[G] => Lambda[G, Event, Result]
+  )(implicit lift: LiftTrace[F, G]): Lambda[F, Event, Result] = { (event, context) =>
     val kernel = HasKernel[Event].extract(event)
     entryPoint.continueOrElseRoot(context.functionName, kernel).use { span =>
-      lift.run(span)(lambda(_)(event, context.mapK(lift.liftK)))
+      lift.run(span) { trace =>
+        trace.put(
+          Tags.arn(context.invokedFunctionArn),
+          Tags.requestId(context.awsRequestId)
+        ) >> lambda(trace)(event, context.mapK(lift.liftK))
```
armanbilge, line 37: Generally I prefer composability to hard-coding, but this doesn't seem unreasonable?

armanbilge: Except, should this be starting a new span or something? I don't really have experience with tracing 😕

bpholt: `continueOrElseRoot` starts a new span inside the root span from which it's continuing, so I think this should be ok.

armanbilge: Thanks, good to know!

lambda-natchez/src/main/scala/feral/lambda/natchez/TracedLambda.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  def apply[F[_]: MonadCancelThrow, G[_], Event: HasKernel, Result](entryPoint: EntryPoint[F])(
+      lambda: Trace[G] => Lambda[G, Event, Result]
+  )(implicit lift: LiftTrace[F, G]): Lambda[F, Event, Result] =
+    viaSpan(entryPoint) { span => (event, context) =>
+      lift.run(span)(lambda(_)(event, context.mapK(lift.liftK)))
+    }
+
+  def viaSpan[F[_]: MonadCancelThrow, Event: HasKernel, Result](entryPoint: EntryPoint[F])(
+      lambda: Span[F] => Lambda[F, Event, Result]): Lambda[F, Event, Result] = {
```
armanbilge, line 36: @bpholt Here's an alternative `TracedLambda` middleware, which only requires a single effect `F` and gives you a `Span[F]` instead of a `Trace[G]`.

Does this mean less boilerplate? It depends. If all you wanted all along was a `Span[F]`, then this is perfect.

But, if you actually want a `Trace` instance (e.g., to use natchez http4s middleware), then you have a couple options.
1. Hardcode the traced effect (either `Kleisli` or `IO`) and run it with the given span. This is straightforward enough, but you lose some abstraction power.
2. Re-invent `LiftTrace[F, G]` 😉

bpholt: This works pretty well! Let's run with it.

armanbilge: Cool! So you prefer this one to the other in terms of `Trace[G]`? I think I'd like to keep both forms, but maybe this should be the primary :)

bpholt:
I think this is in a pretty good state, pending what happens with https://github.com/tpolecat/natchez/pull/448. Is there anything we want to try or do before merging (and maybe publishing snapshots)?

bpholt:
Also, I have the XRay middleware updated to use this branch, but it doesn't build since the XRay support in Natchez hasn't been published yet. I was thinking if we plan to merge this branch soon, I can push it up as a followup PR (in its own module? I'm not sure how we want to organize things yet)

lambda-natchez/src/main/scala/feral/lambda/natchez/TracedLambda.scala
bpholt: I'm using it wrapped by `XRayTracedLambda`:

```scala
package feral.lambda.natchez.xray

import _root_.natchez._
import _root_.natchez.noop.NoopSpan
import _root_.natchez.xray.{XRay, XRayEnvironment}
import cats.data._
import cats.effect.std.Random
import cats.effect.{Trace => _, _}
import feral.lambda._
import feral.lambda.natchez._

object XRayTracedLambda {
  def apply[F[_]: Async, Event: HasKernel, Result](installer: Resource[Kleisli[F, Span[F], *], Lambda[Kleisli[F, Span[F], *], Event, Result]])
                                                  (implicit LT: LiftTrace[F, Kleisli[F, Span[F], *]]): Resource[F, Lambda[F, Event, Result]] =
    installer.mapK(Kleisli.applyK(new NoopSpan[F]())).flatMap(XRayTracedLambda(_))

  def apply[F[_]: Async, Event: HasKernel, Result](lambda: Span[F] => Lambda[F, Event, Result]): Resource[F, Lambda[F, Event, Result]] =
    Resource.eval(Random.scalaUtilRandom[F]).flatMap { implicit random =>
      Resource
        .eval(XRayEnvironment[F].daemonAddress)
        .flatMap {
          case Some(addr) => XRay.entryPoint[F](addr)
          case None => XRay.entryPoint[F]()
        }
        .map(TracedLambda.viaSpan(_)(lambda))
    }

  def apply[F[_]: Async, G[_], Event: HasKernel, Result](lambda: Lambda[G, Event, Result])
                                                        (implicit LT: LiftTrace[F, G]): Resource[F, Lambda[F, Event, Result]] =
    XRayTracedLambda { span => (event, context) =>
      LT.run(span) { _ => lambda(event, context.mapK(LT.liftK)) }
    }
}
```

Here's an example of a Lambda using that middleware: https://github.com/Dwolla/rabbitmq-topology-backup/blob/xray/src/main/scala/com/dwolla/rabbitmq/topology/LambdaHandler.scala#L24-L42

bpholt: Also, here's a custom CFN resource using it: https://github.com/Dwolla/postgresql-init-custom-resource/blob/feral/db/src/main/scala/com/dwolla/postgres/init/Handler.scala#L33-L52

armanbilge:
Awesome! Actually, what I'd like to do is split the PRs again which got mixed up while we were experimenting :)
1. A PR introducing the effect-agnostic `Lambda`, which hopefully @djspiewak can take a look at, since it's fairly different from his [original concept gist](https://gist.github.com/djspiewak/37a4ea0d7a5237144ec8b56a76ed080d) and also since he was dubious about my last attempt at effect agnosticism :)
2. A PR introducing the `TracedLambda`, at least with the `Span[F]`-based formulation since we can do this with Natchez as published today. If `LiftTrace` becomes a published thing, we can consider adding the `Trace[G]`-based formulation as well.
3. A couple more tweak PRs with the `INothing` thing and maybe some other minor optimizations/fixes.

All of this should be publishable right now and I'm happy to do so while Daniel looks at (1).

> Also, I have the XRay middleware updated to use this branch, but it doesn't build since the XRay support in Natchez hasn't been published yet.

Ooh, I'll take a look at this shortly! I thought we could get to a place where we don't need a specific integration, but maybe this was too ambitious 😆 

> in its own module? I'm not sure how we want to organize things yet

I'm like wayyy into over-modularization, so please push back on this. But yes that's my first instinct!

armanbilge:
@bpholt I just published everything in this branch, propagating now.
```
"com.armanbilge" %%% "feral-lambda-natchez" % "0.1-f497734" // and friends :)
```

armanbilge:
Everything here now split into:
- https://github.com/typelevel/feral/pull/51
- https://github.com/typelevel/feral/pull/55
- https://github.com/typelevel/feral/pull/52

PR #52, opened 2021-11-19
Use `Nothing` for no `Result`

armanbilge:

@bpholt Before cherry-picking 00eb18f I wanted to try and write a test in 06c9eeec49322099cebb3934c7c37596932b575f that demonstrates the compiler issue with `Nothing`. I tried replicating your example from https://github.com/typelevel/feral/pull/45#discussion_r751496835 but didn't have much luck. When you have chance would you mind poking at this? Thanks :)


bpholt:
This is the best minimization I've found so far:

```scala
class LambdaSuite {
  def lambda[F[_]]: Resource[F, Lambda[F, Unit, Nothing]] = ???

  def middleware[F[_], Event, Result, Env](lambda: Resource[F, Lambda[Kleisli[F, Env, *], Event, Result]]): Resource[F, Lambda[F, Event, Result]] = ???

  implicit val nothing: Encoder[Nothing] = _ => ???

  class NothingLambda extends IOLambda[Unit, Nothing] {
    def handler =
      middleware(lambda[Kleisli[IO, Unit, *]].mapK(Kleisli.applyK(())))
  }
}
```

```
[error] …/feral/lambda/shared/src/test/scala/feral/lambda/LambdaSuite.scala:35:14: type mismatch;
[error]  found   : cats.effect.kernel.Resource[[+A]cats.effect.IO[A],feral.lambda.Lambda[[γ$1$]cats.data.Kleisli[cats.effect.IO,Unit,γ$1$],Unit,Nothing]]
[error]     (which expands to)  cats.effect.kernel.Resource[[+A]cats.effect.IO[A],(Unit, feral.lambda.Context[[γ$1$]cats.data.Kleisli[cats.effect.IO,Unit,γ$1$]]) => cats.data.Kleisli[cats.effect.IO,Unit,Option[Nothing]]]
[error]  required: cats.effect.kernel.Resource[[+A]cats.effect.IO[A],feral.lambda.Lambda[[γ$0$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$0$],Unit,Result]]
[error]     (which expands to)  cats.effect.kernel.Resource[[+A]cats.effect.IO[A],(Unit, feral.lambda.Context[[γ$0$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$0$]]) => cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,Option[Result]]]
[error]         .mapK(Kleisli.applyK(())))
[error]              ^
[error] one error found
```

It doesn't matter if the value passed to `middleware` is a `Resource` in `F[_]` or `Kleisli[F, E, *]`. If it's the latter, then the `mapK` in the `NothingLambda` goes away, but AFAICT it doesn't change the inference failure:

```
[error] …/feral/lambda/shared/src/test/scala/feral/lambda/LambdaSuite.scala:33:24: type mismatch;
[error]  found   : cats.effect.kernel.Resource[[γ$2$]cats.data.Kleisli[cats.effect.IO,Unit,γ$2$],feral.lambda.Lambda[[γ$2$]cats.data.Kleisli[cats.effect.IO,Unit,γ$2$],Unit,Nothing]]
[error]     (which expands to)  cats.effect.kernel.Resource[[γ$2$]cats.data.Kleisli[cats.effect.IO,Unit,γ$2$],(Unit, feral.lambda.Context[[γ$2$]cats.data.Kleisli[cats.effect.IO,Unit,γ$2$]]) => cats.data.Kleisli[cats.effect.IO,Unit,Option[Nothing]]]
[error]  required: cats.effect.kernel.Resource[[γ$1$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$1$],feral.lambda.Lambda[[γ$0$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$0$],Unit,Result]]
[error]     (which expands to)  cats.effect.kernel.Resource[[γ$1$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$1$],(Unit, feral.lambda.Context[[γ$0$]cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,γ$0$]]) => cats.data.Kleisli[[+A]cats.effect.IO[A],Unit,Option[Result]]]
[error]       middleware(lambda[Kleisli[IO, Unit, *]])
[error]                        ^
[error] one error found
```

armanbilge:
@bpholt thanks! That's an excellent minimization, exactly what I was looking for :)

Unfortunately, placing the `INothing` encoder in the companion object for `IOLambda` doesn't seem to bring it in scope for whatever reason. I'm going to futz with this some more.

armanbilge:
Darn, 2.12 seems to be lacking. Dare I say ... we drop it? 😆

bpholt:
Dropping 2.12 seems reasonable to me. I don't think the common reasons why people are stuck on 2.12 would apply in this context.

armanbilge:
Yeah, apparently we are "future-forward" or something 😆 

If it becomes an issue we can probably bring back 2.12 very easily anyway. Those users would just need to explicitly import this implicit and we'd need to make this test 2.13/3.x-only. FTR if sbt supported this config out-of-the-box that's exactly what I would have done.

armanbilge:
Ummm ... Scala 3 is broken too? Seems like a compiler bug 🤔 
```
[error] -- Error: /workspace/feral/lambda/shared/src/test/scala/feral/lambda/LambdaSuite.scala:35:54 
[error] 35 |  class NothingLambda extends IOLambda[Unit, INothing] {
[error]    |                                                      ^
[error]    |ambiguous implicit arguments: both feral.lambda.nothingEncoder and feral.lambda.nothingEncoder match type io.circe.Encoder[feral.lambda.INothing] of parameter encoder of constructor IOLambda in class IOLambda
```

kubukoz:
tl;dr but I like the idea of `Nothing` in the result :D

armanbilge:
@kubukoz you're not stuck on Scala 2.12 are you? The tl;dr here is mostly about implicit scope for `Nothing` not working on 2.12.

kubukoz:
I'm on 2.13, `Nothing` was giving me a missing `Decoder` instance error.

armanbilge:
Try `INothing`.

kubukoz:
man, I can only have one experimental branch in my dependencies at a time 😂

lambda/shared/src/main/scala-3/feral/lambda/INothing.scala
```diff
@@ -0,0 +1,38 @@
...
+object INothing {
+
+  /**
+   * This can't actually be used. It's here because `IOLambda` demands an Encoder for its result
+   * type, which should be `Nothing` when no output is desired. Userland code will return an
+   * `Option[Nothing]` which is only inhabited by `None`, and the encoder is only used when the
+   * userland code returns `Some`.
+   */
+  @nowarn("msg=dead code following this construct")
+  implicit val nothingEncoder: Encoder[INothing] = (_: INothing) => ???
```
kubukoz, line 37: not 100% sure but can't this just return its argument? `(a: INothing) => a`. `INothing <: Nothing` and `Nothing <: Json`.

armanbilge:
Actually you should have this in your branch, it's split off :)

kubukoz:
we're reaching levels of stable that shouldn't be possible!

lambda/shared/src/main/scala-3/feral/lambda/INothing.scala
armanbilge: Good idea, this does work :)

lambda/shared/src/test/scala/feral/lambda/LambdaSuite.scala
```diff
@@ -0,0 +1,41 @@
...
+class LambdaSuite {
+
+  // Test for INothing type inference
+
+  def lambda[F[_]]: Lambda[F, Unit, INothing] = ???
+
+  def middleware[F[_], Event, Result](lambda: Lambda[Kleisli[F, Unit, *], Event, Result])
+      : Resource[F, Lambda[F, Event, Result]] = ???
+
+  class NothingLambda extends IOLambda[Unit, INothing] {
```
kubukoz, line 35: if I'm not mistaken this should even work with just `Nothing`, at least on 2.13

kubukoz:
This seems mergeable as well...

armanbilge:
Yep, I'm just dealing with the conflicts rn :)

PR #54, opened 2021-11-19
Lambda linting etc.

armanbilge:

### DIFF
--- a/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -21,10 +21,12 @@ import cats.effect.IO
 import cats.effect.kernel.Resource
 import com.amazonaws.services.lambda.{runtime => lambdaRuntime}
 import io.circe
+import io.circe.Printer
+import io.circe.syntax._
 
 import java.io.InputStream
 import java.io.OutputStream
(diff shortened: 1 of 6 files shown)


lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -45,10 +47,15 @@ private[lambda] abstract class IOLambdaPlatform[Event, Result]
...
-          _ <- OptionT(apply(event, context, setup)).foldF(IO.unit) { result =>
-            // TODO can circe write directly to output?
-            IO(output.write(encoder(result).noSpaces.getBytes(StandardCharsets.UTF_8)))
+          context <- IO(Context.fromJava[IO](context))
+          _ <- OptionT(lambda(event, context)).foldF(IO.unit) { result =>
+            IO {
+              val json = result.asJson
+              val bb = Printer.noSpaces.printToByteBuffer(json)
+              val ch = Channels.newChannel(output)
+              ch.write(bb)
```
armanbilge, line 56: I opened https://github.com/circe/circe/pull/1855 which would let us write directly to the `OutputStream`, but not sure what's going on there rn.

bpholt: Looks like there are new maintainers, but they probably need to get settled in :)

PR #55, opened 2021-11-19
Natchez integration for Lambda

armanbilge:


lambda-natchez/src/main/scala/feral/lambda/natchez/HasKernel.scala
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.lambda.natchez
+
+import feral.lambda.events.ApiGatewayProxyEventV2
+import natchez.Kernel
+
+trait HasKernel[Event] {
```
armanbilge, line 22: 🚲 🏠

lambda-natchez/src/main/scala/feral/lambda/natchez/Tags.scala
```diff
@@ -0,0 +1,25 @@
...
+
+package feral.lambda.natchez
+
+import natchez.TraceValue
+
+object Tags {
+  private[this] val prefix = "aws"
+  def arn(s: String): (String, TraceValue) = s"$prefix.arn" -> s
+  def requestId(s: String): (String, TraceValue) = s"$prefix.requestId" -> s
+}
```
armanbilge, line 25: IDK what I'm doing, does this do what we want? Any other tags to consider?

bpholt: This seems like a good start. I don't have enough experience using tags like this to anticipate wanting anything else.

lambda-natchez/src/main/scala/feral/lambda/natchez/Kernel.scala
```diff
@@ -0,0 +1,55 @@
...
+  def seed(result: Result, kernel: Kernel): Result
+}
+
+object KernelSink {
+  @inline def apply[R](implicit ev: KernelSink[R]): ev.type = ev
+
+  implicit val apiGateProxyResultV2KernelSink: KernelSink[ApiGatewayProxyResultV2] = {
+    case (result: ApiGatewayProxyStructuredResultV2, kernel) =>
+      result.copy(headers = kernel.toHeaders ++ result.headers)
+  }
```
armanbilge, line 51: Oh wait, is this not a thing? I guess tracing headers don't really belong in a response 🙃

lambda-natchez/src/main/scala/feral/lambda/natchez/TracedLambda.scala
```diff
@@ -0,0 +1,39 @@
...
+import cats.effect.kernel.MonadCancelThrow
+import cats.syntax.all._
+import feral.lambda.Lambda
+import natchez.EntryPoint
+import natchez.Span
+
+object TracedLambda {
+
+  def apply[F[_]: MonadCancelThrow, Event: KernelSource, Result](entryPoint: EntryPoint[F])(
+      lambda: Span[F] => Lambda[F, Event, Result]): Lambda[F, Event, Result] = {
```
djspiewak, line 28: Given a parametric `Lambda`, wouldn't the better type here be something like `Lambda[Kleisli[F, Span[F], *], Event, Result]`? Could also bring in Cats MTL and do an `Ask[F, Span[F]]`, though that starts to get odd.

armanbilge: Not necessarily. natchez also supports tracing with `IOLocal` instead of Kleisli which IIUC is also more performant. I played with a few other ideas in https://github.com/typelevel/feral/pull/50. Also hoping that https://github.com/tpolecat/natchez/pull/448 becomes a thing.

armanbilge: Oh btw, I now have an MTL lambda concept in https://github.com/typelevel/feral/pull/60. Actually this is my favorite so far, and I think http4s will be moving in this direction as well.

armanbilge:
Integrated with https://github.com/typelevel/feral/pull/60, so it will merge in from there.

PR #68, opened 2021-12-09
Add DynamoDBStreamEvent event type

kubukoz:


lambda-events/src/main/scala/feral/lambda/events/DynamoDBStreamEvent.scala
```diff
@@ -0,0 +1,103 @@
...
+
+final case class AttributeValue(
+    b: Option[String],
+    bs: Option[String],
+    bool: Option[Boolean],
+    l: Option[List[AttributeValue]],
+    m: Option[Map[String, AttributeValue]],
+    n: Option[String],
+    ns: Option[List[String]],
+    nul: Option[true],
```
armanbilge, line 30: We're going to drop 2.12 so this should be fine. Literal types yay!

kubukoz: tbh this could be `nul: Boolean`, but the decoder will be nastier :(

armanbilge:
@kubukoz do you want to update this now that we've dropped 2.12 in main?

kubukoz:
That should cut it

armanbilge (approved):
LGTM thanks for the contribution!

PR #77, opened 2021-12-13
Remove circe-generic

### LINKED ISSUE #76: Remove circe-generic
Assign to me :)

kubukoz:
Closes #76.

### DIFF
--- a/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/package.scala
+++ b/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/package.scala
@@ -19,7 +19,6 @@ package feral.lambda
 import cats.syntax.all._
 import io.circe._
 import io.circe.syntax._
-import io.circe.generic.semiauto._
 import monix.newtypes._
 import org.http4s.Uri
 import org.http4s.circe.CirceInstances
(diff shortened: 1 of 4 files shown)


lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/package.scala
```diff
@@ -94,7 +93,7 @@ package cloudformation {
     }
   }
 
-  case class CloudFormationCustomResourceRequest[A](
+  final case class CloudFormationCustomResourceRequest[A](
       RequestType: CloudFormationRequestType,
```
kubukoz, line 97: IMO these should be `camelCase`, but I didn't want to tackle that in this PR.

armanbilge: Yeah this is @bpholt's baby, I promised him not to mess with it 😝

bpholt: Camel case is fine; IIRC it's only this way to make the decoder easier.

kubukoz: gotchu, let's follow up in another PR

PR #78, opened 2021-12-13
Create `IOLambda.Simple`

armanbilge:
This restores Daniel's original concept. Also good for simple use-cases and those who aren't into the MTL stuff.

### DIFF
--- a/examples/src/main/scala/feral/examples/KinesisLambda.scala
+++ b/examples/src/main/scala/feral/examples/KinesisLambda.scala
@@ -20,12 +20,7 @@ import cats.effect._
 import feral.lambda._
 import feral.lambda.events.KinesisStreamEvent
 
-object KinesisLambda extends IOLambda[KinesisStreamEvent, INothing] {
-  val handler: Resource[IO, LambdaEnv[IO, KinesisStreamEvent] => IO[Option[Nothing]]] =
-    Resource.pure { env =>
-      env
-        .event
-        .flatMap { event => IO.println(s"Received event with ${event.records.size} records") }
-        .as(None)
-    }
+object KinesisLambda extends IOLambda.Simple[KinesisStreamEvent, INothing] {
+  def handle(event: KinesisStreamEvent, context: Context[IO], init: Init) =
+    IO.println(s"Received event with ${event.records.size} records").as(None)
 }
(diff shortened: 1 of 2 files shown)


examples/src/main/scala/feral/examples/KinesisLambda.scala
```diff
@@ -20,12 +20,7 @@ import cats.effect._
...
-    Resource.pure { env =>
-      env
-        .event
-        .flatMap { event => IO.println(s"Received event with ${event.records.size} records") }
-        .as(None)
-    }
+object KinesisLambda extends IOLambda.Simple[KinesisStreamEvent, INothing] {
+  def handle(event: KinesisStreamEvent, context: Context[IO], init: Init) =
+    IO.println(s"Received event with ${event.records.size} records").as(None)
 }
```
armanbilge, line 26: @kubukoz wdyt?

lambda/shared/src/main/scala/feral/lambda/IOLambda.scala
```diff
@@ -43,3 +43,26 @@ abstract class IOLambda[Event, Result](
...
+
+object IOLambda {
+
+  abstract class Simple[Event, Result](
+      implicit decoder: Decoder[Event],
+      encoder: Encoder[Result])
+      extends IOLambda[Event, Result] {
+
+    type Init
+    def init: Resource[IO, Init] = Resource.pure(null.asInstanceOf[Init])
```
armanbilge, line 55: `Init` instead of `Setup`, because `Setup` is already used higher up in the inheritance hierarchy.

kubukoz (approved):

examples/src/main/scala/feral/examples/KinesisLambda.scala
kubukoz: looks good :)

lambda/shared/src/main/scala/feral/lambda/IOLambda.scala
kubukoz: `Env`? `Environment`?

armanbilge: Hmm. That seems more confusing. "Environment" or "Context" is usually something out of your control I think, whereas `Init` is any userland initialization.

kubukoz: reasonable, Init it is then

PR #79, opened 2021-12-13
Tiny refactor: minimize `if`

kubukoz:


armanbilge (approved):
👍 I wanted to try this but I was lazy :P thanks!

Issue #57, opened 2021-11-19
Lambda modules organization?


armanbilge:
This keeps coming up, e.g. https://github.com/typelevel/feral/pull/2, https://github.com/typelevel/feral/pull/6#discussion_r733085235, https://github.com/typelevel/feral/issues/48, https://github.com/typelevel/feral/pull/50#issuecomment-973647953. So at some point we should decide what we want to do :)

Just some assorted thoughts, all IMHO:
* One reason to keep `lambda-events` separate from `lambda` is b/c a shapeless dependency (historically) was a big deal. Not sure if that's true anymore, and irrelevant to Scala 3 or if we hand-write our encoders/decoders.
* It's a bit weird if some events are in `lambda-events` and others are with specialized lambdas. But also, these models aren't really _shared_ interfaces, just kind of helpers ... if e.g. the CloudFormation lambda wants to define its own version of its events and eschew this dependency, I think that could be fine.
* Besides defining events, it could be nice to define aliases for `Lambda` that pair events with results. This idea makes me consider if we should just merge `lambda-events` with `lambda`.
* I'm open to merging `lambda-natchez` into `lambda` so we can offer tracing middleware out-of-the-box. `natchez-core` is lightweight.
* I still don't _think_ we need a specific XRay integration, but could totally be wrong.

armanbilge:
@bpholt and I chatted on [Discord](https://discord.com/channels/632277896739946517/918373380003082250/919772971378278412). Here's what we've decided for now:
1. Include the natchez integration in the core lambda module. "Tracing is the future, let’s bake it in." - @bpholt 
2. Merge the events module into the core lambda module as well. This is more likely to be temporary, just to make it easier to try out feral without needing to add multiple dependencies. Some more wise words from @bpholt:
  > I prefer keeping the models with the specializations because I’ve been frustrated digging thru the dozens of AWS models when using other libraries, looking for the specific ones that were applicable to my use case. Since theoretically people will already be pulling in a specific artifact for the specialized lambda, IMO they might as well get the specialized model too.

kubukoz:
IMO: natchez in core is fine, events in core is fine as long as there are no extra dependencies.

armanbilge:
Yep, I'll make sure to drop circe-generic.

kubukoz:
covered in #77

armanbilge:
I think the current organization is good.

PR #83, opened 2021-12-13
Bake tracing into lambda

armanbilge:
Executing https://github.com/typelevel/feral/issues/57#issuecomment-992056584. Also takes the opportunity to clean up the tracing integration a bit.


lambda/shared/src/main/scala/feral/lambda/TracedLambda.scala
```diff
@@ -40,4 +43,18 @@ object TracedLambda {
     }
   } yield result
 
+  def apply[Event, Result](entryPoint: EntryPoint[IO])(lambda: Trace[IO] => IO[Option[Result]])(
+      implicit env: LambdaEnv[IO, Event],
+      KS: KernelSource[Event]): IO[Option[Result]] = for {
```
armanbilge, line 48: I'm a bit concerned this will be ambiguous with the other constructor. Let me add some tests.

PR #88, opened 2021-12-14
Update the README

armanbilge:
Any and all feedback appreciated!


README.md
```diff
@@ -1,15 +1,55 @@
...
+libraryDependencies += "org.typelevel" %%% "feral-lambda" % "0.1.0-M1"
+
+// Specialized integrations
+libraryDependencies += "org.typelevel" %%% "feral-lambda-http4s" % "0.1.0-M1"
+libraryDependencies += "org.typelevel" %%% "feral-lambda-cloudformation-custom-resource" % "0.1.0-M1"
 ```
+
+`IOLambda[Event, Result]` or `IOLambda.Simple[Event, Result]` are the entrypoint for your Lambda (comparable to `IOApp` in Cats Effect). `Event` and `Result` can be any type for which a circe `Decoder` and `Encoder` are defined and must correspond to the "trigger" event and (optional) result for your Lambda. The `feral.lambda.events` package provides some of these models already ([please consider contributing!](https://github.com/typelevel/feral/issues/48)).
+
+The [examples](https://github.com/typelevel/feral/tree/main/examples/src/main/scala/feral/examples) provide an in-depth demonstration of how to implement your Lambda.
```
kubukoz, line 20: Not really in-depth at this moment :)

kubukoz (approved):

README.md
armanbilge: Haha, good point. I'm planning to take a pass through and flesh them out a bit, but even after that I suspect this will be an exaggeration :P

ChristopherDavenport (approved):

PR #89, opened 2021-12-14
Netlify http4s handler

### LINKED ISSUE #74: Netlify integration
https://docs.netlify.com/functions/build-with-typescript/

kubukoz:
Might close #74 in the future.

image

Instructions for testing:

1. get netlify-cli (e.g. `nix shell nixpkgs#netlify-cli`)
2. run `sbt fastOptJS`
3. `cp netlify/.js/target/scala-2.13/netlify-fastopt/main.js netlify/functions/example.js` (`netlify/functions` is the default directory for functions)
4. run `netlify functions:serve --debug`
5. send curls to `localhost:9999/.netlify/functions/example`


build.sbt
```diff
@@ -141,8 +142,18 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
...
-  .in(file("examples"))
-  .dependsOn(lambda)
-  .enablePlugins(NoPublishPlugin)
+lazy val netlify =
+  crossProject(JSPlatform)
+    .crossType(CrossType.Pure)
+    .in(file("netlify"))
+    .dependsOn(lambdaHttp4s)
+    .settings(
+      scalaJSUseMainModuleInitializer := true
```
kubukoz, line 151: remove after example is removed

build.sbt
```diff
@@ -141,8 +142,18 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
...
+    )
+
+lazy val examples =
```
kubukoz, line 154: fix formatting so this doesn't show as a diff

lambda/js/src/main/scala/feral/lambda/ContextPlatform.scala
```diff
@@ -27,14 +27,14 @@ private[lambda] trait ContextCompanionPlatform {
       context.functionName,
```
kubukoz, line 27: Netlify will need its own context type.

lambda/js/src/main/scala/feral/lambda/ContextPlatform.scala
```diff
@@ -27,14 +27,14 @@ private[lambda] trait ContextCompanionPlatform {
...
       context.functionVersion,
       context.invokedFunctionArn,
-      context.memoryLimitInMB.toInt,
```
kubukoz, line 30: Netlify seems to pass an int here, but the facade uses a string.

netlify/src/main/scala/feral/netlify/NetlifyHttp4sHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+
+object NetlifyHttp4sHandler {
+
+  def apply[F[_]: Concurrent](
+      routes: HttpRoutes[F]
+  )(implicit env: LambdaEnv[F, NetlifyHttpEvent]): F[Option[NetlifyHttpResult]] =
+    for {
+      event <- env.event
+      method <- Method.fromString(event.httpMethod).liftTo[F]
+      uri <- Uri.fromString(event.path).liftTo[F]
```
kubukoz, line 40: Currently this is the relative path. Is that correct or should it be the full URL?

netlify/src/main/scala/feral/netlify/NetlifyHttp4sHandler.scala
```diff
@@ -0,0 +1,74 @@
...
+      headers = Headers(event.headers.toList)
```
kubukoz, line 41: this whole file is nearly identical to the AWS lambda. Some abstraction would be great to have

netlify/src/main/scala/feral/netlify/NetlifyHttpEvent.scala
```diff
@@ -0,0 +1,65 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.netlify
+import io.circe.Decoder
+import cats.data.NonEmptyList
+
+final case class NetlifyHttpEvent(
```
kubukoz, line 21: accurate based on typescript model in https://github.com/netlify/functions/blob/86697792ef363211f367cbf2fab5334389a739ed/src/function/event.ts

netlify/src/main/scala/feral/netlify/NetlifyHttpEvent.scala
```diff
@@ -0,0 +1,65 @@
...
+    "queryStringParameters",
+    "multiValueQueryStringParameters",
+    "body",
+    "isBase64Encoded"
+  )(NetlifyHttpEvent.apply)
+}
+
+import io.circe.Encoder
+
+final case class NetlifyHttpResult(
```
kubukoz, line 51: this is a copy-paste from the Lambda code, I'll review it with netlify's types later. It seems to work so far

netlify/src/main/scala/feral/netlify/handler.scala
```diff
@@ -0,0 +1,39 @@
...
+
+package feral.netlify
+
+import cats.effect.IO
+import cats.effect.kernel.Resource
+import feral.lambda._
+import org.http4s.HttpRoutes
+import org.http4s.Response
+
+object handler extends IOLambda[NetlifyHttpEvent, NetlifyHttpResult] {
```
kubukoz, line 25: this is demo code

build.sbt
```diff
@@ -146,3 +147,17 @@ lazy val examples = crossProject(JSPlatform, JVMPlatform)
...
   .dependsOn(lambda)
   .enablePlugins(NoPublishPlugin)
+
+lazy val netlifyFunctions =
+  crossProject(JSPlatform)
+    .crossType(CrossType.Pure)
+    .in(file("netlify-functions"))
+    .dependsOn(core)
+    .settings(
+      scalaJSUseMainModuleInitializer := true,
```
armanbilge, line 157:
```suggestion

lazy val netlifyFunctions = project
    .in(file("netlify-functions"))
    .enablePlugins(ScalaJSPlugin)
    .dependsOn(core.js)
    .settings(
      name := "feral-netlify-functions"
      scalaJSUseMainModuleInitializer := true,
```

netlify-functions/src/main/scala/feral/netlify/facade/Context.scala
```diff
@@ -0,0 +1,35 @@
...
+  def callbackWaitsForEmptyEventLoop: Boolean = js.native
+  def functionName: String = js.native
+  def functionVersion: String = js.native
+  def invokedFunctionArn: String = js.native
+  def memoryLimitInMB: Int = js.native
+  def awsRequestId: String = js.native
+  def logGroupName: String = js.native
+  def logStreamName: String = js.native
+  def identity: UndefOr[Null | js.Object] = js.native
+  def clientContext: UndefOr[Null | js.Object] = js.native
```
armanbilge, line 33:
```suggestion
  def identity: UndefOr[js.Object] = js.native
  def clientContext: UndefOr[js.Object] = js.native
```

See https://github.com/scala-js/scala-js-dom/issues/481#issuecomment-898192431

netlify-functions/src/main/scala/feral/netlify/LambdaEnv.scala
```diff
@@ -0,0 +1,68 @@
...
+    new LambdaEnv[G, Event] {
+      def event = f(outer.event)
+      def context = f(outer.context).map(_.mapK(f))
+    }
+}
+
+object LambdaEnv {
+  def apply[F[_], A](implicit env: LambdaEnv[F, A]): LambdaEnv[F, A] = env
+
+  implicit def kleisliLambdaEnv[F[_]: Functor, A, B](
```
armanbilge, line 45: I'm 👍  to adding an MTL dependency in a follow-up PR to vanish this bloat.

netlify-functions/src/main/scala/feral/netlify/LambdaEnv.scala
```diff
@@ -0,0 +1,68 @@
...
+import cats.data.Kleisli
+import cats.data.OptionT
+import cats.data.WriterT
+import cats.effect.IO
+import cats.effect.IOLocal
+import cats.kernel.Monoid
+import cats.syntax.all._
+import cats.~>
+
+sealed trait LambdaEnv[F[_], Event] { outer =>
```
armanbilge, line 31: I think "Lambda" is just an AWS thing. Maybe rename this `NetlifyEnv` or `FunctionsEnv` or similar?

netlify-functions/src/main/scala/feral/netlify/IOLambda.scala
```diff
@@ -0,0 +1,88 @@
...
+        context <- IOLocal(context)
+        env = LambdaEnv.ioLambdaEnv(event, context)
+        result <- handler(env)
+      } yield result
+    }
+
+  def main(args: Array[String]): Unit = {
+    val handlerName = getClass.getSimpleName.init
+    js.Dynamic.global.exports.updateDynamic(handlerName)(handlerFn)
+  }
```
armanbilge, line 51: So IIUC _all_ netlify serverless functions have to export a method named `handler`. So I wonder if we can just simplify the design here and use an idiomatic `@JSExport`.

netlify-functions/src/main/scala/feral/netlify/facade/Context.scala
kubukoz: hmm so that'll just be a null, I see

netlify-functions/src/main/scala/feral/netlify/IOLambda.scala
kubukoz: I don't see why not, that makes the most sense.

kubukoz: guess we can't do a top level export in an abstract class... I've changed this to use a constant string but I don't know if we can do much more

netlify-functions/src/main/scala/feral/netlify/IOFunction.scala
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+  def main(args: Array[String]): Unit = {
+    // Netlify functions require the entrypoint to be called `handler`
+    js.Dynamic.global.exports.updateDynamic("handler")(handlerFn)
+  }
+
+  private lazy val handlerFn
+      : js.Function2[js.Any, facade.Context, js.Promise[js.Any | Unit]] = {
+    (event: js.Any, context: facade.Context) =>
```
armanbilge, line 55:
```suggestion
  @JSExportTopLevel("handler")
  def exportedHandler(event: js.Any, context: facade.Context): js.Promise[js.Any | Unit]] =
```

kubukoz: "Only static objects may export their members to the top level"

netlify-functions/src/main/scala/feral/netlify/LambdaEnv.scala
kubukoz: #96, we'll see which lands first

netlify-functions/src/main/scala/feral/netlify/IOFunction.scala
armanbilge: Dang. Thanks for trying.

netlify-functions/src/main/scala/feral/netlify/IOLambda.scala
armanbilge: Oh sorry, I missed this comment before.

build.sbt
```diff
@@ -146,3 +147,17 @@ lazy val examples = crossProject(JSPlatform, JVMPlatform)
...
+lazy val netlifyFunctions =
+  project
+    .in(file("netlify-functions"))
+    .enablePlugins(ScalaJSPlugin)
+    .dependsOn(core.js)
+    .settings(
+      name := "feral-netlify-functions",
+      libraryDependencies ++= Seq(
+        "org.http4s" %%% "http4s-server" % http4sVersion,
+        "io.circe" %%% "circe-core" % circeVersion,
```
armanbilge, line 160: `circe-scalajs` gets us `circe-core` anyway :)
```suggestion
```

netlify-functions/src/main/scala/feral/netlify/NetlifyHttp4sHandler.scala
```diff
@@ -0,0 +1,73 @@
...
+object NetlifyHttp4sHandler {
+
+  def apply[F[_]: Concurrent](
+      routes: HttpRoutes[F]
+  )(implicit env: FunctionEnv[F, HttpFunctionEvent]): F[Option[HttpFunctionResult]] =
+    for {
+      event <- env.event
+      method <- Method.fromString(event.httpMethod).liftTo[F]
+      uri <- Uri.fromString(event.path).liftTo[F]
+      headers = Headers(event.headers.toList)
```
armanbilge, line 40: So this problem also applies to the AWS lambda which is admittedly half-baked, do we need to include `multiValueHeaders` here as well? Also the `rawQuery`.

netlify-functions/src/main/scala/feral/netlify/facade/Context.scala
```diff
@@ -0,0 +1,35 @@
...
+import scala.scalajs.js
+import scala.scalajs.js._
+
+@js.native
+private[netlify] sealed trait Context extends js.Object {
+  def callbackWaitsForEmptyEventLoop: Boolean = js.native
+  def functionName: String = js.native
+  def functionVersion: String = js.native
+  def invokedFunctionArn: String = js.native
+  def memoryLimitInMB: Int = js.native
```
armanbilge, line 28: This is a `String` in the netlify TS, any ideas about that?
https://github.com/netlify/functions/blob/main/src/function/context.ts#L10

armanbilge (commented):
IIUC, the only kind of netlify function possible is an http handler? In which case, I think we can bake this in much more. E.g.

- No need for `Event` and `Result` types or `Option` for the response or implicit decoders/encoders. We can prescribe exactly what these are going to be.
- We could even have `IOFunction.Simple` be implemented directly with an `HttpRoutes`.

build.sbt
kubukoz: true, I thought an explicit dep wouldn't hurt

netlify-functions/src/main/scala/feral/netlify/NetlifyHttp4sHandler.scala
kubukoz: probably so, yeah. I'll check and add them

netlify-functions/src/main/scala/feral/netlify/facade/Context.scala
kubukoz: yeah I saw that, but if you try running it locally you'll get a number. I suspect TS doesn't care as there's no runtime, and coerces the value to a string, but scalajs sees a number and freaks out?

armanbilge: Cool, so it's a mistake with the TS. Good enough for me.

kubukoz: Reported upstream https://github.com/netlify/functions/pull/251

kubukoz:
> IIUC, the only kind of netlify function possible is an http handler? In which case, I think we can bake this in much more.

There's more - you can have functions triggered by deployment events etc. https://docs.netlify.com/functions/trigger-on-events/

technically, it seems like these have the same interface but they have a common body schema, except there's a "background function" type that might not have to return...

armanbilge:
Huh, interesting. As you point out it seems that the http Event/Response are the common shared interface (idk about the background one, but what's the harm in returning a 204 or something lol). Then (a future PR can) add models for those additional triggers and maybe some convenience builders based on them.

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -31,7 +31,7 @@ private[lambda] trait IOLambdaPlatform[Event, Result] {
     js.Dynamic.global.exports.updateDynamic(handlerName)(handlerFn)
   }
 
-  private lazy val handlerFn
+  protected lazy val handlerFn
```
kubukoz, line 34: todo: fix this after https://github.com/typelevel/feral/pull/116 lands

netlify-functions/src/main/scala/feral/netlify/NetlifyHttp4sHandler.scala
kubukoz: still not done, todo

netlify-functions/src/main/scala/feral/netlify/example.scala
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.netlify
+
+import cats.effect.IO
+import feral.lambda.Context
+
+object example extends IOFunction.Simple {
```
kubukoz, line 22: to be removed later

lambda/js/src/main/scala/feral/lambda/ContextPlatform.scala
armanbilge: Follow-up: turns out not.

lambda/js/src/main/scala/feral/lambda/ContextPlatform.scala
armanbilge: Follow-up: inconsistency in the simulator/emulator, we'll support both.

lambda/js/src/main/scala/feral/lambda/facade/Context.scala
```diff
@@ -17,14 +17,17 @@
 package feral.lambda.facade
 
 import scala.scalajs.js
+import scala.scalajs.js._
```
armanbilge, line 20: nitpick :P
```suggestion
import scala.scalajs.js.|
```

lambda/js/src/main/scala/feral/lambda/ContextPlatform.scala
```diff
@@ -27,7 +27,10 @@ private[lambda] trait ContextCompanionPlatform {
...
+      (context.memoryLimitInMB: Any) match {
+        case s: String => s.toInt
+        case i: Int => i
```
armanbilge, line 32: Hmm, why isn't fatal warnings complaining about this 😆

netlify-functions/src/main/scala/feral/netlify/HttpFunctionEvent.scala
```diff
@@ -0,0 +1,65 @@
...
+ * limitations under the License.
+ */
+
+package feral.netlify
+
+import cats.data.NonEmptyList
+import io.circe.Decoder
+import io.circe.Encoder
+
+final case class HttpFunctionEvent(
```
armanbilge, line 23: Let's add a test using this example:
https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

netlify-functions/src/main/scala/feral/netlify/IOFunction.scala
```diff
@@ -0,0 +1,56 @@
...
+  override def main(args: Array[String]): Unit = {
+    // Netlify functions require the entrypoint to be called `handler`
+    js.Dynamic.global.exports.updateDynamic("handler")(handlerFn)
+  }
+
+}
+
+object IOFunction {
+
+  abstract class Simple extends IOFunction {
```
armanbilge, line 37: Can this extend `IOLambda.Simple` instead and avoid the repetition?

netlify-functions/src/main/scala/feral/netlify/example.scala
armanbilge: Worth moving to examples?

netlify-functions/src/main/scala/feral/netlify/HttpFunctionEvent.scala
armanbilge: Oh also let's add a `KernelSource` for this so the tracing middleware works too.

lambda/js/src/main/scala/feral/lambda/facade/Context.scala
```diff
@@ -17,14 +17,17 @@
...
 
 @js.native
 private[lambda] sealed trait Context extends js.Object {
   def callbackWaitsForEmptyEventLoop: Boolean = js.native
   def functionName: String = js.native
   def functionVersion: String = js.native
   def invokedFunctionArn: String = js.native
-  def memoryLimitInMB: String = js.native
+  // Note: This should always be a string, but the Netlify CLI currently passes a number.
+  // ref: https://github.com/netlify/functions/pull/251
```
armanbilge, line 29: Upstreamed again to https://github.com/ashiina/lambda-local/pull/218 but I'm not optimistic.

armanbilge:
So apparently a netlify `Context` looks like this:
```javascript
{
  callbackWaitsForEmptyEventLoop: [Getter/Setter],
  succeed: [Function (anonymous)],
  fail: [Function (anonymous)],
  done: [Function (anonymous)],
  functionVersion: '$LATEST',
  functionName: '164d0769d4ca9c18ce26562dca3f9e408070c2d1c2b884621e91527e1724e77d',
  memoryLimitInMB: '1024',
  logGroupName: '/aws/lambda/164d0769d4ca9c18ce26562dca3f9e408070c2d1c2b884621e91527e1724e77d',
  logStreamName: '2021/08/03/[$LATEST]d3dabe643c01435bb985e7b0ae5821d0',
  clientContext: {
    custom: {
      netlify: 'eyJpZGVudGl0eSI6eyJ1cmwiOiJodHRwczovL2Fkb3JpbmctYWxiYXR0YW5pLWEyODkwNi5uZXRsaWZ5LmFwcC8ubmV0bGlmeS9pZGVudGl0eSIsInRva2VuIjoiZXlKaGJHY2lPaUpJVXpJMU5pSXNJblI1Y0NJNklrcFhWQ0o5LmV5SmxlSEFpT2pFMk1qYzVOekkxTVRRc0luTjFZaUk2SWpBaWZRLlJGSDVwRDlPUGZUa2lSb3RscXdRekJmbU1JWlF0QWFOVGNkYlVnX2N6WWMifSwidXNlciI6eyJhcHBfbWV0YWRhdGEiOnsicHJvdmlkZXIiOiJlbWFpbCJ9LCJlbWFpbCI6InRlc3RAdGVzdC5jb20iLCJleHAiOjE2Mjc5NzU5NjgsInN1YiI6Ijc3NWE1MzQ0LTBjNmQtNGRjNi05YWM3LWQyZDY0NmUwNzFlYyIsInVzZXJfbWV0YWRhdGEiOnsiZnVsbF9uYW1lIjoiVGVzdCBVc2VyIn19LCJzaXRlX3VybCI6Imh0dHBzOi8vYWRvcmluZy1hbGJhdHRhbmktYTI4OTA2Lm5ldGxpZnkuYXBwIn0='
    },
    identity: {
      url: 'https://adoring-albattani-a28906.netlify.app/.netlify/identity',
      token: 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2Mjc5NzI1MTQsInN1YiI6IjAifQ.RFH5pD9OPfTkiRotlqwQzBfmMIZQtAaNTcdbUg_czYc'
    },
    user: {
      app_metadata: [Object],
      email: 'test@test.com',
      exp: 1627975968,
      sub: '775a5344-0c6d-4dc6-9ac7-d2d646e071ec',
      user_metadata: [Object]
    }
  },
  identity: undefined,
  invokedFunctionArn: 'arn:aws:lambda:us-east-1:375708248955:function:164d0769d4ca9c18ce26562dca3f9e408070c2d1c2b884621e91527e1724e77d',
  awsRequestId: '2ee47065-afe2-433f-b88d-3280e25b5ae4',
  getRemainingTimeInMillis: [Function: getRemainingTimeInMillis]
}
```
via https://answers.netlify.com/t/could-not-access-user-from-context-inside-netlify-function/41801/6

The bad news is that there does seem to be some custom stuff in `ClientContext` in addition to the actual `custom` field being used. So maybe there is some post-processing going on 🤔 . _Also_, I'm not even sure if those things are "POJSOs" since they rendered as `[Object]` which suggests it's an instance of a class i.e. not something we can run through circe.

This is really annoying. I don't want to give up all the nice sharing though so I think have to think what to do about it.

lambda/js/src/main/scala/feral/lambda/facade/Context.scala
armanbilge: Fixed and released!

armanbilge:
Some additional notes on `clientContext`:
1. `custom.netlify` will be available via https://github.com/typelevel/feral/pull/117.
2. The `user` object is actually just the payload of the JWT carried in the `Authorization: Bearer token` header. So, we can get at it ourselves without modifying our `Context` facade.
3. The `identity` object, it's not clear to me if we can get at it another way—it would be great to experiment with Netlify and see how if we can figure out how it's injected into the request (e.g. a header, environment variable, etc.). Otherwise, I'm inclined to ignore it until somebody complains 😝

kubukoz:
clearly lost interest in this, I'll close for now but anyone can feel free to take whatever they want from this :)

PR #90, opened 2021-12-14
Fix dynamodb stream event decoder

kubukoz:
This slipped through the cracks somehow...


armanbilge (approved):
Maybe someday we can have tests :)

kubukoz:
yeah, but end to end 😅

Issue #85, opened 2021-12-13
Is `LambdaEnv` powerful enough?


armanbilge:
The `LambdaEnv` was born in #60 to help the implementation of lambda middlewares. But notably, `LambdaEnv` doesn't actually enable transformation of the `Event` or `Context`, like an http4s middleware does: it is read-only. This suits our purposes perfectly fine so far, but I am curious if there is some legit need to actually transform the `Event` before passing it on for further processing.

bpholt:
One could think of the CloudFormation custom resource lambda as a kind of middleware that takes events of type `CloudFormationCustomResourceRequest[Input]` and gives them to a handler of type `Input => F[HandlerResponse[Output]]`. 

Actually, now that I say that, I wonder if we should change the `CloudFormationCustomResource` trait so that its methods take the `Input` type instead of `CloudFormationCustomResourceRequest[Input]`, e.g.

```scala
trait CloudFormationCustomResource[F[_], Input, Output] {
  def createResource(event: Input): F[HandlerResponse[Output]]
  def updateResource(event: Input): F[HandlerResponse[Output]]
  def deleteResource(event: Input): F[HandlerResponse[Output]]
}
```

The full request is available from the implicit `LambdaEnv`, so implementations would still have access to everything, and in my experience, that `Input` type is what you're really interested in anyway.

(I was actually thinking the other day that it was curious that we removed the context from those methods but not the request itself. I figured it made sense to still have explicit unwrapped access to the input, so I didn't say anything at the time.)

That isn't directly responsive to your question but maybe it will spark some discussion. 🙂

armanbilge:
Aha, fantastic point :) 

> The full request is available from the implicit LambdaEnv, so implementations would still have access to everything, and in my experience, that Input type is what you're really interested in anyway.

💯 it's pretty much the same with the http4s lambda. Let's run with this!

armanbilge:
Seems to be working fine.

PR #94, opened 2021-12-15
`LambdaEnv` improvements

armanbilge:
Some ideas I had while working on the examples. Comments below.


lambda/shared/src/main/scala/feral/lambda/LambdaEnv.scala
```diff
@@ -42,6 +42,15 @@ sealed trait LambdaEnv[F[_], Event] { outer =>
 object LambdaEnv {
   def apply[F[_], A](implicit env: LambdaEnv[F, A]): LambdaEnv[F, A] = env
 
+  def of[F[_]: Applicative, Event](e: Event, c: Context[F]): LambdaEnv[F, Event] =
+    new LambdaEnv[F, Event] {
+      val event = e.pure
+      val context = c.pure
+    }
```
armanbilge, line 49: This is a sort of "escape-hatch" that lets you create a `LambdaEnv` if you have an `Event` and `Context`. This makes it possible to use the `LambdaEnv`-based middlewares even from `IOLambda.Simple`.

lambda/shared/src/main/scala/feral/lambda/LambdaEnv.scala
```diff
@@ -42,6 +42,15 @@ sealed trait LambdaEnv[F[_], Event] { outer =>
...
+
+  def event[F[_], Event](implicit env: LambdaEnv[F, Event]): F[Event] = env.event
+  def context[F[_], Event](implicit env: LambdaEnv[F, Event]): F[Context[F]] = env.context
```
armanbilge, line 52: Some convenience methods. Didn't do it here, but might less us improve the UI a bit.

```scala
// before
def myMethod[F[_]: Monad](implicit env: LambdaEnv[F, MyEvent]) = for {
  event <- env.event
} yield ???

// after
// we can provide these aliases for each type of lambda
type MyEventLambdaEnv[F[_]] = LambdaEnv[F, MyEvent]
def myMethod[F[_]: Monad: MyEventLambdaEnv] = for {
  event <- LambdaEnv.event
} yield ???
```

Subject to personal preference, of course. But I think I prefer the latter style.

lambda/shared/src/main/scala/feral/lambda/IOLambda.scala
```diff
@@ -30,15 +31,13 @@ abstract class IOLambda[Event, Result](
...
-        result <- handler(env)
-      } yield result
-    }
+  final override protected def setup: Resource[IO, Setup] = for {
+    handler <- handler
+    localEvent <- IOLocal[Event](null.asInstanceOf[Event]).toResource
+    localContext <- IOLocal[Context[IO]](null).toResource
+    env = LambdaEnv.ioLambdaEnv(localEvent, localContext)
+    result = handler(env)
+  } yield { localEvent.set(_) *> localContext.set(_) *> result }
```
armanbilge, line 40: This looks a bit uglier and is less safe, but all optimizations are 😉 this lets us avoid building the user-provided handler on every invocation, since acquiring the handler resource actually returns a function `LambdaEnv => IO` and not the final `IO`.

kubukoz: is that going to be more efficient if we'll be doing two `.set`s on these locals in every request anyway?

armanbilge: Hmm, why would using `.set` be inefficient?

So if your function `LambdaEnv => IO` is something like:
`env => Middleware(Middleware(Middleware(handler)))`
basically I'm trying to avoid invoking that method on every request.

lambda/shared/src/main/scala/feral/lambda/LambdaEnv.scala
armanbilge: On second thought, adding this API is a bit dangerous. In https://github.com/typelevel/feral/issues/85 I considered whether we may need additional methods on `LambdaEnv` to e.g. allow transformation of the `Event` in some way. In which case, an `Applicative` constraint would no longer be powerful enough to create a `LambdaEnv`.

armanbilge: Reverted.

lambda/shared/src/main/scala/feral/lambda/LambdaEnv.scala
armanbilge: Did it. I do like it :)

PR #96, opened 2021-12-15
Use cats-mtl to derive LambdaEnv instances

kubukoz:


lambda/shared/src/test/scala/feral/lambda/LambdaEnvLiftSuite.scala
```diff
@@ -0,0 +1,39 @@
...
+
+  implicit val env: LambdaEnv[IO, Event]
+
+  implicitly[LambdaEnv[Kleisli[IO, String, *], Event]]
+  implicitly[LambdaEnv[WriterT[IO, String, *], Event]]
+  implicitly[LambdaEnv[StateT[IO, String, *], Event]]
+  implicitly[LambdaEnv[EitherT[IO, String, *], Event]]
+  implicitly[LambdaEnv[OptionT[IO, *], Event]]
+  // doesn't work with this PR, works without it
+  // implicitly[LambdaEnv[Kleisli[EitherT[IO, String, *], String, *], Event]]
```
armanbilge, line 38: Update: this works on Scala 3 with this PR.

armanbilge: Update: nevermind, can't read 😆

kubukoz: I miscommunicated, it doesn't :D

armanbilge:
This has bincompat implications so we have to decide before 0.1.0 :)

kubukoz:
tbh I think it was better before this, at least we could support arbitrary compositions of transformers. This way we only get one level... which is less than ideal.

armanbilge:
It's annoying boilerplate but not the end of the world.

PR #98, opened 2021-12-16
Add tests for events, other improvements/fixes

armanbilge:
Tests are scala-2 only for now because I used circe-literal, but that's been ported to Scala 3 for the next circe release :)


kubukoz:
oh nice, didn't know circe-literal was ported to scala 3

armanbilge:
Yep :) in https://github.com/circe/circe/pull/1874

PR #110, opened 2021-12-20
Switch publishing to s01

armanbilge:
Daniel seemed to think `org.typelevel` may already be setup for `s01.oss.sonatype.org`. So, let's find out.


armanbilge:
Let's see what happens...

PR #116, opened 2021-12-22
Enable JS lambda handler name to be customized

armanbilge:
Exposes the `handlerName` as an overriddable `def`. Keeps the default to the `object` name .

@kubukoz you can hard-code this to "handler" for your Netlify `IOFunction`.


kubukoz:
this is good, I was actually making `handlerFn` a `protected` so that I could override `main`...

kubukoz (approved):

armanbilge:
👍 I figure other people may want to do this too.

kubukoz:
:shipit:

PR #117, opened 2021-12-22
Add support for `custom` field in `ClientContext`

armanbilge:
By modelling it as a `JsonObject` and parsing/decoding it as such.

In fact, it seems the entirety of `ClientContext` is parsed from JSON in the Java and Node.js Lambda runtimes.
- https://github.com/aws/aws-lambda-java-libs/blob/4fe08f417a4f5b612bb8d88006da3341eb6d3e51/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/EventHandlerLoader.java#L109
- https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/c31c41ffe5f2f03ae9e8589b96f3b005e2bb8a4a/src/Runtime/InvokeContext.ts#L114

### DIFF
--- a/lambda/js/src/main/scala/feral/lambda/facade/Context.scala
+++ b/lambda/js/src/main/scala/feral/lambda/facade/Context.scala
@@ -42,7 +42,7 @@ private[lambda] sealed trait CognitoIdentity extends js.Object {
 @js.native
 private[lambda] sealed trait ClientContext extends js.Object {
   def client: ClientContextClient = js.native
-  def Custom: js.Any = js.native
+  def custom: js.UndefOr[js.Any] = js.native
   def env: ClientContextEnv = js.native
 }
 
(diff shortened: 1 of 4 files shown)


lambda/js/src/main/scala/feral/lambda/facade/Context.scala
```diff
@@ -42,7 +42,7 @@ private[lambda] sealed trait CognitoIdentity extends js.Object {
 @js.native
 private[lambda] sealed trait ClientContext extends js.Object {
   def client: ClientContextClient = js.native
-  def Custom: js.Any = js.native
+  def custom: js.UndefOr[js.Any] = js.native
```
armanbilge, line 45: I'm 99% sure this was a typo in the Typescript. I don't know why it would be different for Node.js, or why it would be capitalized in general.

PR #124, opened 2021-12-25
aws-lambda-java-core is a provided dep

armanbilge:
Well, I assume it should be anyway 🤔 can't find any documentation saying this.


armanbilge:
Huh, really can't find anything that says this. I guess I'm wrong.

PR #142, opened 2022-01-08
Migrate to sbt-typelevel

armanbilge:


build.sbt
```diff
@@ -94,32 +68,18 @@ val munitVersion = "0.7.29"
...
-    )
-    .enablePlugins(NoPublishPlugin)
+  tlCrossRootProject.aggregate(
+    core,
+    lambda,
+    sbtLambda,
+    lambdaHttp4s,
+    lambdaCloudFormationCustomResource,
+    examples
+  )
```
armanbilge, line 82: Not only does this cut down the boilerplate, it also gets CI split into JVM and JS jobs for free.

PR #144, opened 2022-01-10
use PUT instead of POST for CloudFormation Custom Resource responses…

bpholt:
… since that's what AWS expects.

From [Custom resource provider response fields](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-responses.html):

> For more information about uploading objects by using presigned URLs, see the related topic in the Amazon Simple Storage Service User Guide.

From [the Amazon Simple Storage Service User Guide](https://docs.aws.amazon.com/AmazonS3/latest/userguide/PresignedUrlUploadObject.html):

> When you create a presigned URL, you must provide your security credentials and then specify a bucket name, an object key, an HTTP method (PUT for uploading objects), and an expiration date and time. The presigned URLs are valid only for the specified duration. That is, you must start the action before the expiration date and time. If the action consists of multiple steps, such as a multipart upload, all steps must be started before the expiration, otherwise you will receive an error when Amazon S3 attempts to start a step with an expired URL.

### DIFF
--- a/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
+++ b/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
@@ -23,7 +23,7 @@ import cats.syntax.all._
 import feral.lambda.cloudformation.CloudFormationRequestType._
 import io.circe._
 import io.circe.syntax._
-import org.http4s.Method.POST
+import org.http4s.Method.PUT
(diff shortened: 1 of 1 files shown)


armanbilge (approved):
On Discord I asked whether `PUT` makes sense for all the event types (create, update, delete) and response URLs. @bpholt explains:
> cloudformation triggers the lambda and then just waits for something to be written to the S3 path it gives you as part of the request
> the response written to S3 has a specific format that varies a bit depending on whether the lambda considers itself to have succeeded or failed

PR #146, opened 2022-01-11
fix some CloudFormation Custom Resource response issues

bpholt:
This drops null values from the responses, and minimizes the stack traces included in failure responses to help make sure the responses stay under 4KB.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -175,7 +175,6 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
       "org.typelevel" %%% "scalacheck-effect" % "1.0.3" % Test,
       "org.typelevel" %%% "scalacheck-effect-munit" % "1.0.3" % Test,
       "com.eed3si9n.expecty" %%% "expecty" % "0.15.4" % Test,
-      "io.circe" %%% "circe-optics" % "0.14.1" % Test,
       "io.circe" %%% "circe-literal" % "0.14.1" % Test,
       "io.circe" %%% "circe-testing" % "0.14.1" % Test
     )
--- a/build.sbt
+++ b/build.sbt
@@ -168,7 +168,16 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
     libraryDependencies ++= Seq(
       "io.monix" %%% "newtypes-core" % "0.0.1",
       "org.http4s" %%% "http4s-client" % http4sVersion,
-      "org.http4s" %%% "http4s-circe" % http4sVersion
+      "org.http4s" %%% "http4s-circe" % http4sVersion,
+      "org.http4s" %%% "http4s-dsl" % http4sVersion % Test,
+      "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
+      "org.typelevel" %%% "munit-cats-effect-3" % "1.0.7" % Test,
+      "org.typelevel" %%% "scalacheck-effect" % "1.0.3" % Test,
+      "org.typelevel" %%% "scalacheck-effect-munit" % "1.0.3" % Test,
+      "com.eed3si9n.expecty" %%% "expecty" % "0.15.4" % Test,
+      "io.circe" %%% "circe-optics" % "0.14.1" % Test,
+      "io.circe" %%% "circe-literal" % "0.14.1" % Test,
+      "io.circe" %%% "circe-testing" % "0.14.1" % Test
     )
   )
   .settings(commonSettings)
--- a/build.sbt
+++ b/build.sbt
@@ -65,6 +65,7 @@ val http4sVersion = "0.23.7"
 val natchezVersion = "0.1.6"
 val munitVersion = "0.7.29"
 val munitCEVersion = "1.0.7"
+val scalacheckEffectVersion = "1.0.3"
 
 lazy val commonSettings = Seq(
   crossScalaVersions := Seq(Scala3, Scala213)
@@ -100,7 +101,7 @@ lazy val lambda = crossProject(JSPlatform, JVMPlatform)
       "io.circe" %%% "circe-scodec" % circeVersion,
       "org.scodec" %%% "scodec-bits" % "1.1.30",
       "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
-      "org.typelevel" %%% "munit-cats-effect-3" % "1.0.7" % Test
+      "org.typelevel" %%% "munit-cats-effect-3" % munitCEVersion % Test
     ),
     libraryDependencies ++= {
       if (tlIsScala3.value) Nil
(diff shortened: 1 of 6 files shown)


lambda-cloudformation-custom-resource/src/test/scala/feral/lambda/cloudformation/CloudFormationCustomResourceArbitraries.scala
```diff
@@ -0,0 +1,201 @@
...
+
+  def genLambdaEnv[F[_]: Applicative, A: Arbitrary]
+      : Gen[LambdaEnv[F, CloudFormationCustomResourceRequest[A]]] =
+    for {
+      e <- arbitrary[CloudFormationCustomResourceRequest[A]]
+      c <- arbitrary[Context[F]]
+    } yield new LambdaEnv[F, CloudFormationCustomResourceRequest[A]] {
+      override def event: F[CloudFormationCustomResourceRequest[A]] = e.pure[F]
+      override def context: F[Context[F]] = c.pure[F]
+    }
```
armanbilge, line 195: Instead of unsealing `LambdaEnv` can we just upstream this constructor? I'd considered adding it before, but hesitated due to https://github.com/typelevel/feral/issues/85. We can keep it package-private.

lambda-cloudformation-custom-resource/src/test/scala/feral/lambda/cloudformation/ResponseSerializationSuite.scala
```diff
@@ -0,0 +1,125 @@
...
+          _ <- CloudFormationCustomResource(client, new DoNothingCustomResource[IO])
+          (req, body) <- eventualRequest.get.timeout(2.seconds)
+          event <- lambdaEnv.event
+        } yield {
+          val expectedJson = event.RequestType match {
+            case OtherRequestType(requestType) =>
+              val expectedReason = s"unexpected CloudFormation request type `$requestType`"
+              val expectedStackTraceHead =
+                s"java.lang.IllegalArgumentException: $expectedReason"
+              json"""{
```
armanbilge, line 69: These tests will have to be Scala 2 only to get the JSON literals I think

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
```diff
@@ -38,6 +39,12 @@ trait CloudFormationCustomResource[F[_], Input, Output] {
 }
 
 object CloudFormationCustomResource {
+  private implicit def jsonEncoder[F[_]]: EntityEncoder[F, Json] =
+    jsonEncoderWithPrinter(
+      Printer(
+        dropNullValues = true,
+        indent = ""
+      ))
```
armanbilge, line 47: WDYT about `defaultPrinter.copy(dropNullValues = true)` instead? So we always whatever good defaults are in the http4s encoder.

Also, I wonder if we need this fix also in the encoder for lambdas that return an output, like the http one.

build.sbt
```diff
@@ -168,7 +168,15 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
...
-      "org.http4s" %%% "http4s-circe" % http4sVersion
+      "org.http4s" %%% "http4s-circe" % http4sVersion,
+      "org.http4s" %%% "http4s-dsl" % http4sVersion % Test,
+      "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
+      "org.typelevel" %%% "munit-cats-effect-3" % "1.0.7" % Test,
+      "org.typelevel" %%% "scalacheck-effect" % "1.0.3" % Test,
+      "org.typelevel" %%% "scalacheck-effect-munit" % "1.0.3" % Test,
+      "com.eed3si9n.expecty" %%% "expecty" % "0.15.4" % Test,
+      "io.circe" %%% "circe-literal" % "0.14.1" % Test,
+      "io.circe" %%% "circe-testing" % "0.14.1" % Test
```
armanbilge, line 179: Just a nit, do you mind condensing these versions at the top? I think I already have some there.

lambda-cloudformation-custom-resource/src/test/scala/feral/lambda/cloudformation/CloudFormationCustomResourceArbitraries.scala
bpholt: Yep, good idea

lambda-cloudformation-custom-resource/src/test/scala/feral/lambda/cloudformation/ResponseSerializationSuite.scala
bpholt: Maybe I should convert them back to use `Json.obj` instead. :-(

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
bpholt: I'm not sure about the other lambdas—it's possible, but it's also possible that the nulls should stay in those. Probably needs to be evaluated for each type. I'll switch to the default printer here though.

build.sbt
bpholt: Sure!

lambda-cloudformation-custom-resource/src/test/scala/feral/lambda/cloudformation/ResponseSerializationSuite.scala
armanbilge: JSON literals are coming in circe 0.15 so I've been just hanging on :)

lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
bpholt: FYI `defaultPrinter` is protected in `CirceInstances` so I'm going to use `Printer.noSpaces` instead.

armanbilge: Oh yeah that's fine. Or inherit from `CirceInstances` too maybe. Whatever works.

build.sbt
```diff
@@ -168,7 +169,14 @@ lazy val lambdaCloudFormationCustomResource = crossProject(JSPlatform, JVMPlatfo
     libraryDependencies ++= Seq(
       "io.monix" %%% "newtypes-core" % "0.0.1",
       "org.http4s" %%% "http4s-client" % http4sVersion,
-      "org.http4s" %%% "http4s-circe" % http4sVersion
+      "org.http4s" %%% "http4s-circe" % http4sVersion,
+      "org.http4s" %%% "http4s-dsl" % http4sVersion % Test,
+      "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
+      "org.typelevel" %%% "munit-cats-effect-3" % "1.0.7" % Test,
```
armanbilge, line 175:
```suggestion
      "org.typelevel" %%% "munit-cats-effect-3" % munitCEVersion % Test,
```

armanbilge (approved):
LGTM. Glad we finally got some tests for this, thanks for that! :)

Do you need a milestone after this merges, or maybe test the snapshot to see if there's any more bugs first?

bpholt:
I'll test the snapshot first, but if it works I think we should go ahead with a milestone.

armanbilge:
Gotta run, but please merge when green 👍

PR #157, opened 2022-01-25
tweak CloudFormationCustomResourceResponse decoder

bpholt:
Right now it doesn't work with the JSON emitted by Feral if the `Data` element is empty, because `null`s are pruned by the JSON printer. This doesn't matter in normal use, but it means the decoder can't be used in some testing scenarios.


armanbilge (approved):
Oh I see. That's kind of annoying. We need a better story for `None` vs `null`.

bpholt:
Well, at least in this case the `null` we're talking about is part of the JSON ADT and not literally `null`, so we're not likely to see NullPointerExceptions from it.

PR #182, opened 2022-03-15
Changed ApiGatewayProxyEventV2 cookies field to optional

carpetscheme:
I was trying out the api-gateway proxy integration and found it was failing due to the `cookies` field in the request not being present by default. I couldn't find anything official but [this Typescript model](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d7d9c1d76dead09380e702d858b40432a29d8a48/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L174) seems to agree. Let me know if you prefer I add some additional tests.

Thanks for the project! I have had really nice results testing it out so far.

### DIFF
--- a/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
+++ b/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
@@ -40,9 +40,7 @@ object ApiGatewayProxyHandler {
       event <- LambdaEnv.event
       method <- Method.fromString(event.requestContext.http.method).liftTo[F]
       uri <- Uri.fromString(event.rawPath + "?" + event.rawQueryString).liftTo[F]
-      cookies = Some(event.cookies)
-        .filter(_.nonEmpty)
-        .map(Cookie.name.toString -> _.mkString("; "))
+      cookies = event.cookies.filter(_.nonEmpty).map(Cookie.name.toString -> _.mkString("; "))
       headers = Headers(cookies.toList ::: event.headers.toList)
       readBody =
         if (event.isBase64Encoded)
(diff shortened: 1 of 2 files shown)


armanbilge:
Good catch! Sure, if you have an example of a failing JSON input you can add to the suite that would be awesome.

https://github.com/typelevel/feral/blob/ad3740edea1ef3d7d4c8e4131408889f9d3ee0ce/lambda/shared/src/test/scala-2/feral/lambda/events/ApiGatewayProxyEventV2Suite.scala

lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
```diff
@@ -80,9 +78,7 @@ object ApiGatewayProxyHandler {
       event: ApiGatewayProxyEventV2): F[Request[F]] = for {
     method <- Method.fromString(event.requestContext.http.method).liftTo[F]
     uri <- Uri.fromString(event.rawPath + "?" + event.rawQueryString).liftTo[F]
-    cookies = Some(event.cookies)
-      .filter(_.nonEmpty)
-      .map(Cookie.name.toString -> _.mkString("; "))
+    cookies = event.cookies.filter(_.nonEmpty).map(Cookie.name.toString -> _.mkString("; "))
```
armanbilge, line 81: Hum, obviously not your fault but I can't remember why the code is duplicated here. Seems like this method was extracted out but then forgot to use it 😅

carpetscheme: Yes I was wondering too 😅 So I can remove the duplication from the apply method and use the extracted method? (Or remove `decodeEvent`, but it is useful for testing)

armanbilge: Yes please, I think the intention was to use `decodeEvent` in `apply` :)

armanbilge (approved):
Excellent! Thank you very much for tidying things up and adding tests.

I'll cut an M6 after I merge.

PR #195, opened 2022-04-07
Add SNSEvent, update to circe-* 0.15.0-M1, use jawn on all platforms

jhenahan:
I'd like to use Feral for a project, but wouldn't you know it! The very first event type I need to respond to isn't set up, yet.

I had to bring in `circe-parser` to be able to deal with `String.Array` properly, which as I understand it obviates the need for `circe-jawn` since `circe-parser` is Jawn on the JVM and `JSON.parse` on ScalaJS. Since the dependency was only actually used once on the JVM side and I needed the more general one, anyway, it seemed reasonable to make that change. Happy to find a better way if that'd be preferred, though.

For SnsMessageAttributes, we're lenient with acceptable attributes but strict in the interpretation of documented types. E.g., `String.Array` will not accept any values that can't be parsed as `Boolean`, `BigDecimal`, or `String`, in that order.

This is a strictly stronger contract than what's enforced by, say, [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/566941a83b5a1b3bed3ec9c8cede12fd3ebdb513/types/aws-lambda/trigger/sns.d.ts) or [aws-lambda-go](https://github.com/aws/aws-lambda-go/blob/bc1ec47cb1670c0d5eaca47c10d89789d8507c3d/events/sns.go), but it aligns with [the documented properties](https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html). Happy to loosen things up if that'd be preferred, but it was natural enough once I got over the `String.Array` hump.

### DIFF
--- a/lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
@@ -19,7 +19,7 @@ package events
 
 import cats.syntax.all._
 import io.circe.Decoder
-import io.circe.parser.decode
+import io.circe.jawn.decode
 import io.circe.scodec._
 import scodec.bits.ByteVector
 
--- a/lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package feral.lambda
 package events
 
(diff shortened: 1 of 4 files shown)


lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
```diff
@@ -0,0 +1,223 @@
...
+    signatureVersion: String,
+    timestamp: Instant,
+    signingCertUrl: String,
+    message: String,
+    unsubscribeUrl: String,
+    subject: String
+)
+
+object SnsMessage {
+  implicit val instantDecoder: Decoder[Instant] = Decoder.decodeString.emapTry { str =>
```
armanbilge, line 136: Is this not in Circe? I think they have date/time instances.

lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
```diff
@@ -0,0 +1,223 @@
...
+        "UnsubscribeUrl": "TEST",
+        "Subject": "Test Message"
+      }
+    }
+  ]
+}
+
+==>
+
+SnsEvent(
```
armanbilge, line 65: Since you've written it out, maybe we check that it parses to this in the test? :)

lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
```diff
@@ -0,0 +1,223 @@
...
+
+    /*
+    See https://docs.aws.amazon.com/sns/latest/dg/sns-message-attributes.html
+
+    In particular, for both Number and String.Array:
+
+    > This data type isn't supported for AWS Lambda subscriptions. If you specify this data type for Lambda endpoints,
+    > it's passed as the String data type in the JSON payload that Amazon SNS delivers to Lambda."
+
+    Circe is smart enough to handle this for BigDecimal, but no such luck for this particular mess.
```
armanbilge, line 182: Oh I see now. Yeah this is all very messy 🙃

armanbilge (commented):
Hi, that's great, thank you very much for the PR! Overall looks great :)

So interesting that you bring up the Circe/Jawn stuff. As as of Circe 0.14.2 Jawn will be supported on _both_ JVM and JS platforms (and is currently available in 0.15.0-M1).

I've flipped-flopped in the past trying to figure out what to do here (see https://github.com/http4s/http4s-dom/issues/10, https://github.com/http4s/http4s/pull/5342), but I think I've settled that using Jawn on both JVM and JS is a good thing.

The main reasons being that it guarantees that parsing behaves the same across both platforms, and furthermore Jawn semantics are _much_ better than the JS-native parser e.g. handling of numerics outside the usual ranges. But I don't know if those semantics matter here.

Also I haven't benchmarked it, and for sure the JS-native parser should be more performant than using Jawn.js (right?) but in practice I also don't really know how much that matters.

Are you using feral on JVM or JS, what are your thoughts on this?

PS didn't check but CI is probably complaining about some annoying stuff, you can run `prePR` in sbt to fix those.

lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
jhenahan: Quite correct. Fixed.

lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
jhenahan: Done!

jhenahan:
> As as of Circe 0.14.2 Jawn will be supported on both JVM and JS platforms (and is currently available in 0.15.0-M1).

I saw that as I was trying to figure out what to do here, actually. I didn't know support would be landing in 0.14.2, so I was mostly just trying to avoid the potential mess of depending on an `M1` and simultaneously not upset ScalaJS compat.

I definitely agree in principle that consistent and correct parsing is probably more important than the fastest possible parsing, especially in the context of Lambdas. I'm happy to take your lead on how to resolve while the version support is up in the air.

> Are you using feral on JVM or JS, what are your thoughts on this?

For the work I want to pick up, I'm looking at building for the JVM just because I have a few hard dependencies that probably won't get ScalaJS support any time soon (largely around data formats like Parquet), but I'm interested in playing with the JS side once I've got some work that'll play nice.

jhenahan:
Pushed the version bump and the necessary changes on the JVM side. I don't know the ScalaJS side well enough to know if anything needs to happen there, but I'm happy to make additional changes if you can give me a pointer.

lambda/shared/src/test/scala-2/feral/lambda/events/SnsEventSuite.scala
```diff
@@ -0,0 +1,122 @@
...
+import scodec.bits.ByteVector
+
+import java.time.Instant
+import java.util.UUID
+
+class SnsEventSuite extends FunSuite {
+  import SnsEventSuite._
+
+  test("decoder") {
+    assert(clue(event.as[SnsEvent].toTry.get) == clue(decoded))
```
armanbilge, line 31: Nit and/or question :) I've never used `clue` before so TIL, but I'd guess that `assertEquals` does effectively the same thing here?

```suggestion
    assertEquals(event.as[SnsEvent].toTry.get, decoded)
```

lambda/shared/src/main/scala/feral/lambda/events/SnsEvent.scala
```diff
@@ -0,0 +1,153 @@
...
+    signatureVersion: String,
+    timestamp: Instant,
+    signingCertUrl: String,
+    message: String,
+    unsubscribeUrl: String,
+    subject: String
+)
+
+object SnsMessage {
+  implicit val instantDecoder: Decoder[Instant] = Decoder.decodeInstant
```
armanbilge, line 69:
```suggestion
  private[this] implicit val instantDecoder: Decoder[Instant] = Decoder.decodeInstant
```

armanbilge (approved):
Looks great, thanks for the contribution! I'll cut an M7 after merging :)

PR #200, opened 2022-04-22
Make subject optional in SNS Event

jhenahan:
A fact about buried in the docs (and not evidently reflected in any other implementation since I guess people love `null`) is that [`Subject` is optional](https://docs.aws.amazon.com/sns/latest/dg/sns-message-and-json-formats.html#http-notification-json)! That's a miss on my part.


armanbilge (approved):
Aha! No worries and thanks for the fix!

armanbilge:
Look for M8 on Maven in 30 minutes or so :)
https://github.com/typelevel/feral/releases/tag/v0.1.0-M8

PR #214, opened 2022-05-26
Establish `series/1.x` against http4s 1.0.0 milestones

armanbilge:
As if we're not already neck-deep in our own milestones :)

Why start `series/1.x` instead of `series/0.2` for these milestones? Because we may need to bump to 0.2.0 before http4s 1.0.0 finalizes, and it would be awkward to rollback to 0.23. Meanwhile, we cannot 1.0.0 ourselves until all of our dependencies 1.0.0 including http4s, so it is sufficiently in the future.

We should really avoid doing any feature development on this branch. At the moment it exists solely to track http4s 1.0.0 milestones.

Overall changes:
1. Bump to http4s 1.0.0-M33.
2. Fixes/optimizations in the lambda-http4s integration module, to account for the new http4s entity model.
3. Drop natchez-http4s from our examples, because it is not tracking the http4s 1.0.0 milestones.

### DIFF
--- a/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
+++ b/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
@@ -27,6 +27,7 @@ import org.http4s.Charset
 import org.http4s.Entity
 import org.http4s.Headers
 import org.http4s.HttpRoutes
+import org.http4s.MalformedMessageBodyFailure
 import org.http4s.Method
 import org.http4s.Request
 import org.http4s.Response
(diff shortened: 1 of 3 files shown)


lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
```diff
@@ -39,12 +43,21 @@ object ApiGatewayProxyHandler {
...
       isBase64Encoded = !response.charset.contains(Charset.`UTF-8`)
-      responseBody <- (if (isBase64Encoded)
-                         response.body.through(fs2.text.base64.encode)
-                       else
-                         response.body.through(fs2.text.utf8.decode)).compile.foldMonoid
+      responseBody <- response.entity match {
+        case Entity.Empty => "".pure
+        case Entity.Strict(chunk) => // TODO
+          if (isBase64Encoded) chunk.toByteVector.toBase64.pure
+          else chunk.toByteVector.decodeUtf8.liftTo
```
armanbilge, line 52: This makes me grumpy. I still think `Entity.Strict` should use a `ByteVector` instead of a `Chunk`.

Issue #226, opened 2022-06-22
Lambda output empty with 1.0.0-M1


skirsdeda:
First of all, this is a great project! And I love the name :)

When using 1.0.0-M1 I get empty output from lambda (Scala 2.13 compiled for JVM 11). Tried downgrading to 0.1.0-M8 and that one works perfectly fine. Could it be that this commit broke the output?: da29dc72c033c2e37716f58a55786f977352ceb1

armanbilge:
Thanks for the report. Do you have the same issue in `0.1.0-M9`?

armanbilge:
Thanks for confirming! I'll take a closer look a bit later :)

PR #227, opened 2022-06-22
Close the output writer in JVM lambda runtime

armanbilge:
Silly mistake. But now we have a test!

Fixes https://github.com/typelevel/feral/issues/226

### DIFF
--- a/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -53,6 +53,7 @@ private[lambda] abstract class IOLambdaPlatform[Event, Result]
               val json = result.asJson
               val writer = new OutputStreamWriter(output)
               Printer.noSpaces.unsafePrintToAppendable(json, writer)
+              writer.close()
             }
           }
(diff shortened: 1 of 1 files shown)


lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -53,6 +53,7 @@ private[lambda] abstract class IOLambdaPlatform[Event, Result]
               val json = result.asJson
               val writer = new OutputStreamWriter(output)
               Printer.noSpaces.unsafePrintToAppendable(json, writer)
+              writer.close()
```
bpholt, line 56: Should this be wrapped in a `Resource` so it's closed even if an exception occurs?

armanbilge: Hmm, good point. Maybe. I didn't reach for `Resource` here because I was stupidly focused on optimizing into a single `IO { ... }` 😅 

I'm not sure if it makes a difference in practice, since if we get an error at this stage the Lambda has effectively crashed since there is nothing to catch it. But I do think it's worth making this code idiomatic.

PR #231, opened 2022-06-23
Add S3BatchEvent and S3BatchResult

skirsdeda:
S3 batch ops event/result types. Cheers!

### DIFF
--- a/lambda/shared/src/main/scala-2/feral/lambda/package.scala
+++ b/lambda/shared/src/main/scala-2/feral/lambda/package.scala
@@ -41,5 +41,7 @@ package object lambda {
   type ApiGatewayProxyLambdaEnv[F[_]] = LambdaEnv[F, ApiGatewayProxyEventV2]
   type DynamoDbStreamLambdaEnv[F[_]] = LambdaEnv[F, DynamoDbStreamEvent]
   type KinesisStreamLambdaEnv[F[_]] = LambdaEnv[F, KinesisStreamEvent]
+  type S3BatchLambdaEnv[F[_]] = LambdaEnv[F, S3BatchEvent]
+  type SnsLambdaEnv[F[_]] = LambdaEnv[F, SnsEvent]
   type SqsLambdaEnv[F[_]] = LambdaEnv[F, SqsEvent]
 }
--- a/lambda/shared/src/main/scala/feral/lambda/events/S3BatchEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/S3BatchEvent.scala
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2022 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.lambda.events
+
+import feral.lambda.KernelSource
+import io.circe.Decoder
+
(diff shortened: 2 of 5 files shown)


armanbilge:
Thanks for the contribution! Looks like this needs `sbt headerCreateAll`, sorry 😅

skirsdeda:
Doesn't like 2022 copyright year, I guess 🙁

lambda/shared/src/main/scala/feral/lambda/events/S3BatchEvent.scala
```diff
@@ -0,0 +1,54 @@
...
+
+object S3BatchEvent {
+  implicit val decoder: Decoder[S3BatchEvent] =
+    Decoder.forProduct4("invocationSchemaVersion", "invocationId", "job", "tasks")(
+      S3BatchEvent.apply)
+
+  implicit def kernelSource: KernelSource[S3BatchEvent] = KernelSource.emptyKernelSource
+}
+
+final case class S3BatchJob(id: String)
```
armanbilge, line 39: Seems it's called `S3BatchEventJob` in TS :)
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d18a30127146217038c61060b06e96713daa706c/types/aws-lambda/trigger/s3-batch.d.ts#L17

lambda/shared/src/main/scala/feral/lambda/events/S3BatchResult.scala
```diff
@@ -0,0 +1,63 @@
...
+  implicit val encoder: Encoder[S3BatchResult] =
+    Encoder.forProduct4(
+      "invocationSchemaVersion",
+      "treatMissingKeysAs",
+      "invocationId",
+      "results")(r =>
+      (r.invocationSchemaVersion, r.treatMissingKeysAs, r.invocationId, r.results))
+}
+
+sealed trait S3BatchResultCode
```
armanbilge, line 40: Admittedly, it's kind of a crazy name 😂 

```suggestion
sealed abstract class S3BatchResultResultCode
```
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d18a30127146217038c61060b06e96713daa706c/types/aws-lambda/trigger/s3-batch.d.ts#L35

lambda/shared/src/main/scala-2/feral/lambda/package.scala
```diff
@@ -41,5 +41,7 @@ package object lambda {
   type ApiGatewayProxyLambdaEnv[F[_]] = LambdaEnv[F, ApiGatewayProxyEventV2]
   type DynamoDbStreamLambdaEnv[F[_]] = LambdaEnv[F, DynamoDbStreamEvent]
   type KinesisStreamLambdaEnv[F[_]] = LambdaEnv[F, KinesisStreamEvent]
+  type S3BatchLambdaEnv[F[_]] = LambdaEnv[F, S3BatchEvent]
+  type SnsLambdaEnv[F[_]] = LambdaEnv[F, SnsEvent]
```
armanbilge, line 45: Thanks for adding this missing one! :)

armanbilge (commented):
Thanks! Just a few small nits, otherwise looking good!

lambda/shared/src/main/scala/feral/lambda/events/S3BatchResult.scala
skirsdeda: No problem. Any reason why `abstract class` works better than `trait` here?

armanbilge: Good question, in practice it will probably make no measurable difference. I prefer `abstract class` where possible because admittedly, superstition, 😂 I think it _might_ be more performant to do `isInstanceOf` checks and stuff.

More interestingly, the new `enum` keyword in Scala 3 would let us define it like this as:
```
enum S3BatchResultResultCode:
  case Succeeded, TemporaryFailure, PermanentFailure
```

And behind-the-scenes, that is generating code equivalent to using an `abstract class`. So this approach is binary compatible with that syntax sugar.

skirsdeda: Gotcha, makes sense.

Issue #209, opened 2022-05-20
v0.1.0


armanbilge:
This has been on my mind, and now that Circe 0.14.2 landed we no longer have milestone dependencies of our own :)

I think the feral-lambda core module has been pretty stable. We did break binary-compatibility when fixing some events, and that may generally be annoying going forward depending on how Amazon evolves their APIs.

Things I'd like to do:
- open up the `LambdaEnv` and `Context` APIs, I think I saw @bpholt doing some shenanigans to mock them
- make sure the API can support graceful shutdown of resources, even if we don't implement it yet. h/t @fdietze for reminding me of this issue
- reassure myself that folks are happy with the API 😅 I was reminded this week that the MTL style can be confusing/controversial, see https://github.com/http4s/http4s/pull/4758#issuecomment-1130153431
- review the events module. de-case-classing may be the best for longterm bincompat, but will be a PITA
- anything else?

There's also a milestone with some issues in it, but they're mostly nice-to-haves. The smithy4s integration would still be good if we are okay to tie our binary-compatibility to them.

https://github.com/typelevel/feral/milestone/1

I'm used to working on projects where bincompat is a Big Deal, and it doesn't have to be here yet. I'd like to stay stable if we can, but worst case is 0.2.0 :)

cc @bpholt and @kubukoz

bpholt:
👍  Feral has been working really well for us and going to 0.1.0 seems like a great next step!

There are some minor tweaks to the CloudFormation custom resource interface that I've been meaning to propose. I'll try to get that written up soon.

armanbilge:
Another thing to toss into the mix: possibly a way to support alternative runtimes. I'm hopeful Scala Native support will land in the near future, which will necessitate implementing our own lambda runtime. The interesting thing is said runtime could also cross-compile to JVM and JS, and be an alternative to the official runtimes. There could be some advantages (for example using the WSTP through-and-through, full control of resource lifecycles) so it would be nice to have that flexibility.

armanbilge:
Well, we ended up publishing a v0.2.0 😅 and Scala Native and stuff got pushed back to some future breaking release.

PR #238, opened 2022-07-05
add PhysicalResourceId to CloudFormation Custom Resource interface

bpholt:
`PhysicalResourceId` is "always sent with Update and Delete requests; never sent with Create" ([according to AWS docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requests.html#crpg-ref-request-physicalresourceid)) and is commonly required to actually perform the work needed to do an update or delete of the resource. This PR encodes that contract and makes life a bit easier for developers implementing update/delete, since they won't have to deal with the optionality of the value in the event encoding.

Obviously this is a breaking change to the interface, thus my comment on https://github.com/typelevel/feral/issues/209#issuecomment-1133172755.

### DIFF
--- a/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
+++ b/lambda-cloudformation-custom-resource/src/main/scala/feral/lambda/cloudformation/CloudFormationCustomResource.scala
@@ -30,8 +30,12 @@ import org.http4s.client.dsl.Http4sClientDsl
 
 trait CloudFormationCustomResource[F[_], Input, Output] {
   def createResource(input: Input): F[HandlerResponse[Output]]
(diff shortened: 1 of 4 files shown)


armanbilge (approved):
Yes, this is a nice improvement to the API :)

Actually I think we could have made this change compatibly by deprecating the old methods, and adding default implementations for the new ones that delegate to the old. But no reason to fuss with that :)

bpholt:
> Actually I think we could have made this change compatibly by deprecating the old methods, and adding default implementations for the new ones that delegate to the old. But no reason to fuss with that :)

I thought about that too (actually I implemented it) but I decided against committing it, because it makes the initial experience worse—when you go to implement `CloudFormationCustomResource`, your editor is likely autogenerate stubs for the deprecated methods, not the ones we want people to use going forward.

Issue #233, opened 2022-06-24
natchez integration and lambda init


skirsdeda:
There's a problem when using tracing in combo with resources that are cached (in init or setup part of execution), which is even included in one of the examples: https://github.com/typelevel/feral/blob/main/examples/src/main/scala/feral/examples/KinesisLambda.scala#L52-L53.

### The problem:
1. Actual root span cannot be used in init because it will never be closed. This is even worse if you reuse `Trace[IO]` instance inside request handler code, because all spans there would become children of this never-ending root span. And some natchez backend implementations (like XRay) only send root span together with all children when root span is finished. So you get no recorded spans as a result.
2. If you use fake root span to solve the first problem (smth like this: https://github.com/tpolecat/natchez/issues/566#issue-1278510665), then spans produced during init (like DB connection pool initialization with skunk) are separate from spans produced in the first request (handler code) of that lambda instance.

This means that `TracedHandler` implementation is useless if you need to have `Trace[IO]` during init. In case of skunk, you would want to see SQL query spans inside request spans and for that you need to reuse same instance of `Trace[IO]` for all requests.

### How it should work:
It should be possible to build a single `Trace[IO]` instance for the lifetime of lambda instance. Spans produced during initialization would get recorded as children of root span for the first ever request to that lambda instance.

I hope I managed to describe this clearly 😅. Would be great to work this out and get that example fixed so that it would show how to use skunk and tracing with feral.

armanbilge:
Thanks for writing this up! Indeed, there are still some things to work out with Tracing :)

If I understand correctly, I _think_ you've run into a version of problem (2) as described in https://github.com/typelevel/otel4s/issues/11#issuecomment-1159464303. See also [bayou](https://github.com/armanbilge/bayou/) for some exploratory work at improving the situation.

In the short term, instead of using `IO`-based tracing, perhaps `Kleisli` based tracing would work here?

armanbilge:
@skirsdeda did you have any more luck with this? Any ideas or feedback you have would be very valuable, thanks!

skirsdeda:
@armanbilge I am working on it. I think I am somewhere around half way through at this point. Should have smth to show next week.

PR #240, opened 2022-07-07
[WIP] TracedIOLambda

skirsdeda:
This solves #233, I think 🙂 

As I found out, `IOSetup.setup` needs to get context of the first request to be able to construct `Trace[IO]` resource. And then, all custom resources are memoized separately so that any trace spans produced during init are part of main request span. This should be evident from `TracedIOLambdaSuite`.

This turned out to be quite a lot of code 😓 I am sure this is not the best implementation possible because my FP skills were at the limit 😄 It would probably be nicer if `IOLambda` and `TracedIOLambda` shared more code. Also, there should be `TracedIOLambda.Simple` to match `IOLambda.Simple`.

Leaving as work-in-progress and waiting for feedback.

### DIFF
--- a/core/src/main/scala/feral/IOSetup.scala
+++ b/core/src/main/scala/feral/IOSetup.scala
@@ -20,11 +20,12 @@ import cats.data.OptionT
 import cats.effect.IO
 import cats.effect.kernel.{Deferred, Resource}
 import cats.syntax.all._
+import scala.annotation.nowarn
 
 private[feral] trait IOSetup[Context] {
 
   protected type Setup
-  protected def setup(ctx: Context): Resource[IO, Setup] =
+  protected def setup(@nowarn ctx: Context): Resource[IO, Setup] =
     Resource.pure(null.asInstanceOf[Setup])
 
   private[this] final lazy val deferred = Deferred.unsafe[IO, Either[Throwable, Setup]]


skirsdeda:
The motivation is quite simple: make it possible to use`skunk` with all it's tracing capabilities.
The way `skunk` works is as follows:
1. DB session is created as `Resource[IO, Session[IO]]` and it takes `Trace[IO]` as an implicit param. On acquisition of this resource, DB connection pool is initialized and this results in a "pool" span. Obviously, when using in feral, we want this memoized and only happen on the first request. This part would be equivalent to "init" span in `TracedIOLambdaSuite`.
2. When DB statements are executed, skunk Session produces "query" spans. These should appear in all requests, which means that `Trace[IO]` instance should be a single one and be aware of changing context of subsequent requests. Again, equivalent of "query" span in `TracedIOLambdaSuite`.

As I described in #233, with existing tracing capabilities in feral this just doesn't work. All spans produced by skunk `Session` get "lost".

armanbilge:
Right, thanks. The part that I'm less sure about is why the traces during the initialization should belong specifically to the span of the initial request, instead of a root "initialization" span for the Lambda itself. IIUC, it would mean that the Lambda is unable to initialize until the first request arrives, which may not be desirable.

armanbilge:
No problem, this is tricky stuff and I appreciate your work here! 🚀

PR #247, opened 2022-07-21
Improve mockability and introduce lambda-testkit

armanbilge:
Inline comments incoming. Addresses a point from https://github.com/typelevel/feral/issues/209.

### DIFF
--- a/lambda-testkit/src/main/scala/feral/lambda/testkit/TestContext.scala
+++ b/lambda-testkit/src/main/scala/feral/lambda/testkit/TestContext.scala
@@ -1,0 +1,46 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


lambda/shared/src/main/scala/feral/lambda/LambdaEnv.scala
```diff
@@ -29,7 +29,7 @@ import cats.kernel.Monoid
 import cats.syntax.all._
 import cats.~>
 
-sealed trait LambdaEnv[F[_], Event] { outer =>
+trait LambdaEnv[F[_], Event] { outer =>
```
armanbilge, line 32: I feel like a more appropriate name for this is `LambdaInvocation` but ... that's a mouthful, and it's a bit late in the day for changes :)

lambda-testkit/src/main/scala/feral/lambda/testkit/TestContext.scala
```diff
@@ -0,0 +1,46 @@
...
+ * limitations under the License.
+ */
+
+package feral.lambda
+package testkit
+
+import scala.concurrent.duration.FiniteDuration
+
+object TestContext {
+  def apply[F[_]](
```
armanbilge, line 23: I thought a lot about how to fix this.

I didn't want to make the `Context` constructor public in the core lambda module, because then we wouldn't be able to add new fields to it backwards-compatibly even when AWS can.

OTOH, if the public constructor is in the testkit, should a new required field be added, we can just use a dummy value as a default. This is something I wouldn't want to do in lambda core, but seems reasonable for testing :)

armanbilge:
- https://github.com/typelevel/feral/pull/425

Issue #246, opened 2022-07-18
ES Module support


hugo-vrijswijk:
Native [ES modules](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules) are finally [supported in all current LTS NodeJS versions](https://nodejs.org/api/esm.html) 🎉. More and more Node libraries are moving to ES modules. ES modules can use a top-level import for CommonJS, but CommonJS cannot import ES modules with top-level imports. This means libraries still using CommonJS risk getting left behind in the Node ecosystem.

Feral currently uses a [dynamic export update](https://github.com/typelevel/feral/blob/24a07363cf0ccd51db6d8b2e026cd975f6801fcc/lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala#L30) to export the handler function. From what I could see, this is done to load resources only once instead of every time the lambda is invoked. However, ES modules do not support dynamic export updates. All exports have to be there as `export` statements. The recommended way for the resource use-case is by using [top-level awaits](https://aws.amazon.com/blogs/compute/using-node-js-es-modules-and-top-level-await-in-aws-lambda/). 

To keep Node.js library interop and keep the library modern, it would be ideal for Feral to also support building as a ES Module. I'm not sure if it is possible to instruct Scala.JS to emit different code based on the module setting, or perhaps a different entrypoint should be available for ES Module users.

armanbilge:
Thanks for raising this issue, this is an important point and something to try and address before 0.1.0.

> Feral currently uses a [dynamic export update](https://github.com/typelevel/feral/blob/24a07363cf0ccd51db6d8b2e026cd975f6801fcc/lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala#L30) to export the handler function. From what I could see, this is done to load resources only once instead of every time the lambda is invoked

Indeed I'm not thrilled about the current encoding. Actually I settled on it in desperation for the friendliest UX. Without it, a user would have to manually override a method in their their and add a `@JSExport` annotation. Hopefully this would probably work with ESModules as well, but I have yet to investigate that. If you have any ideas for this I would greatly appreciate it 🙏 

IMO the biggest issue for Scala.js and ESModules is lack of GCC optimization, which makes a huge difference especially for Typelevel libs which have several layers of abstraction.
- https://github.com/scala-js/scala-js/issues/3893

armanbilge:
[...]

PR #248, opened 2022-07-23
(JS) Expose handler function type to allow manual exporting of Lambda handler

hugo-vrijswijk:
Related to #246

Needs some more docs, maybe some further (manual) testing, and maybe a completely different implementation. But I wanted to open this PR to have an example and get the discussion flowing.

With this, you can implement and export your lambda as follows:

```scala
object Lambda extends IOLambda[String, String] {

  override def handler: Resource[IO, LambdaEnv[IO, String] => IO[Option[String]]] =
    Resource
      .eval(allocationCounter.getAndUpdate(_ + 1))
      .as(_.event.map(Some(_)) <* invokeCounter.getAndUpdate(_ + 1))

  @JSExportTopLevel("handler")
  def impl: HandlerFn = handlerFn
}
```


lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -26,12 +26,23 @@ import scala.scalajs.js.|
...
+   *
+   * @example
+   *   {{{
+   *  val handler: HandlerFn = handlerFn
+   *   }}}
+   */
+  final type HandlerFn = js.Function2[js.Any, facade.Context, js.Promise[js.Any | Unit]]
+
   final def main(args: Array[String]): Unit =
     js.Dynamic.global.exports.updateDynamic(handlerName)(handlerFn)
```
armanbilge, line 41: I guess we can remove this now :)

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -26,12 +26,23 @@ import scala.scalajs.js.|
...
 
   protected def handlerName: String = getClass.getSimpleName.init
 
-  private lazy val handlerFn
+  protected lazy val handlerFn
```
armanbilge, line 45: Should be `final` now that we are exposing it.

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -26,12 +26,23 @@ import scala.scalajs.js.|
...
+  /**
+   * Lambda handler. Implement this type with a val and a call to `handlerFn` to export your
+   * handler.
+   *
+   * @example
+   *   {{{
+   *  val handler: HandlerFn = handlerFn
+   *   }}}
+   */
+  final type HandlerFn = js.Function2[js.Any, facade.Context, js.Promise[js.Any | Unit]]
```
armanbilge, line 38: Does this actually work, if `facade.Context` is package-private?

lambda/js/src/test/scala/feral/lambda/ExportedLambdaSuite.scala
```diff
@@ -0,0 +1,76 @@
...
+class CountingIOLambda(allocationCounter: Ref[IO, Int], invokeCounter: Ref[IO, Int])
+    extends IOLambda[String, String] {
+
+  override def handler: Resource[IO, LambdaEnv[IO, String] => IO[Option[String]]] =
+    Resource
+      .eval(allocationCounter.getAndUpdate(_ + 1))
+      .as(_.event.map(Some(_)) <* invokeCounter.getAndUpdate(_ + 1))
+
+  // This would be exported with `@JSExportTopLevel("handler")`
+  def impl: HandlerFn = handlerFn
```
armanbilge, line 61: Since we're not exporting it, what is this actually testing?

To verify the export works, our test should be able to call the exported method without knowing about the `CountingIOLambda` type.

armanbilge (commented):
Thanks for starting this!! If you're really brave I guess ideally we would have some integration tests against the actual AWS JS runtime that verify it works for both CommonJS and ESModules.

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
hugo-vrijswijk: Yes! The type is public, but the types of the `Function2` are not, and I guess that works :)

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
hugo-vrijswijk: I can remove it if you want, but it would be a breaking change for any existing users. It might be nice to still have the old implementation for CommonJS users. Maybe we can also add a warning or clear error when this is called inside ESModule

lambda/js/src/test/scala/feral/lambda/ExportedLambdaSuite.scala
hugo-vrijswijk: Yeah, I suppose this only tests that `impl` calls `handlerFn` which calls `handler`. Which verifies that the handler works when implemented by the class. I don't think you can explicitly call an export inside a unit test. If you have any ideas about how to approach I'd be happy to hear it

hugo-vrijswijk: I've now added a new sbt scripted test to replace this one. It's similar to the existing one, except it tests the exported handler

README.md
```diff
@@ -23,14 +23,44 @@ Next, implement your Lambda. Please refer to the [examples](examples/src/main/sc
...
+    - The handler for your lambda is `index.yourLambdaName`
+3. Change these settings in your `build.sbt`:
+
+    ```scala
+    import org.scalajs.linker.interface.OutputPatterns
+    import _root_.io.circe.syntax._
+
+    scalaJSUseMainModuleInitializer := false
+    scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.ESModule))
+    npmPackageAdditionalNpmConfig := Map("type" -> "module".asJson)
```
armanbilge, line 58: Hopeing we can simplify this to `scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.ESModule))` :)

lambda/js/src/test/scala/feral/lambda/ExportedLambdaSuite.scala
armanbilge: > I don't think you can explicitly call an export inside a unit test. If you have any ideas about how to approach I'd be happy to hear it

For the record, I'm pretty sure you can :) conceptually an "exported" function is equivalent to a function defined in JavaScript land. So we can write a facade for it, and call it, without knowing that it is in fact defined in Scala!

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
armanbilge: I have to think about this. I think I dislike the idea that there are two distinct ways to define JS Lambdas, and "dead" code depending on which you choose (such as `handlerName` for ESModules). I think this would be very confusing. Since your approach words for both, I'd rather buy-in all the way.

README.md
hugo-vrijswijk: Same! With https://github.com/davenverse/sbt-npm-package/pull/16 we can remove this

hugo-vrijswijk:
> If you're really brave I guess ideally we would have some integration tests against the actual AWS JS runtime that verify it works for both CommonJS and ESModules.

I have no idea how I'd even start on this 😅

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
hugo-vrijswijk: Alright, I suppose a breaking change like this is better to do early in the life of a library. If I have your permission, I'll remove the related code

armanbilge:
> I have no idea how I'd even start on this

If you want to do this 😁 start by making a new project with feral and figure out how to run it against SAM locally. Then, write those commands into an sbt task. Then, we make that into a scripted test.

Some of my old work here might be helpful:
https://github.com/ChristopherDavenport/js-test/tree/serverless

Anyway just driving by, sorry, will come back later with another review pass :) thanks for all your work!

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
armanbilge: Exactly. Please go ahead with the change, although maybe we should make this PR the beginning of an v0.2.x series since this is a significant change (ignoring the fact that 0.1.x hasn't materialized a stable release 😅). You can bump the `tlBaseVersion` in the build to signify this.

hugo-vrijswijk: Done. I don't think I've missed anything, but you are welcome to double-check me :)

armanbilge:
I wonder if instead of dealing with SAM it would be easier to work with the Node.js runtime directly.
https://www.npmjs.com/package/aws-lambda-ric
https://github.com/aws/aws-lambda-nodejs-runtime-interface-client

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
armanbilge: Thanks! I'm afraid the example project will also need updating :) To keep it cross-compiling with JVM I think we can add a `@JSExport` stub.
https://github.com/scala-js/scala-js-stubs/

armanbilge:
Cool, I did some investigation. It seems the trick is to install [aws-lambda-ric](https://www.npmjs.com/package/aws-lambda-ric) and and [aws-lambda-rie](https://github.com/aws/aws-lambda-runtime-interface-emulator).

Then we start the lambda like this:
```
$ aws-lambda-rie aws-lambda-ric app.handler
```

where `app.js` contains
```javascript
exports.handler = async (event, context) => {
  return 'Hello World!';
}
```

which can be invoked like:
```
$ curl -XPOST "http://localhost:8080/2015-03-31/functions/function/invocations" -d '{}'
"Hello World!"
```

No pressure to take this on :) but given my previous hiccups with exports I would like to add this integration test just to be sure that everything works.

armanbilge:
Hmm, this is confusing.
- https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/issues/44

armanbilge:
Ah, instead of installing the ric and rie seems we can directly invoke the docker image. https://hub.docker.com/r/amazon/aws-lambda-nodejs

PR #254, opened 2022-08-11
http4s - support proxy handler from HttpApp

bplommer:


armanbilge (approved):
Thanks! For my own education, is this "proxy handler" a specific thing in http4s?

bplommer:
I think it's an AWS thing rather than an http4s thing? But I don't know, `ApiGateWayProxyHandler` was already there and it looks like you wrote it? 🤷

armanbilge:
Ah 😂 I misunderstood your PR title there. I thought maybe there was an http4s "proxy" middleware that worked on `HttpApp` instead of `HttpRoutes`lol. Ok thanks!

PR #257, opened 2022-08-14
Use `sealed abstract case class`es to reduce bincompat surface

armanbilge:
`sealed abstract case class`es don't have `apply`  or `copy` methods, but still have nice `toString`, `hashCode`, and `equals`.

Unfortunately they still generate `unapply` methods, so I've overridden those with package-private implementations to prevent the public one from being generated.

I've made the `apply` methods for events package-private. In ordinary use they should not be necessary, and importantly it protects our bincompat should AWS decide to add more non-optional fields, which technically they can do backwards-compatibly at any time.

Then, we can expose `apply` methods in the testkit like I propose doing for `Context` in https://github.com/typelevel/feral/pull/247. This will enable them to be mocked.

I didn't finish the DynamoDB-related stuff, I think I might like to make some bigger changes there in a followup.


armanbilge:
- https://github.com/typelevel/feral/pull/425

Issue #134, opened 2022-01-02
Custom Lambda runtime?


armanbilge:
Following from https://github.com/typelevel/feral/issues/132, @baccata points out that AWS offers a similarly low level interface to build a custom Lambda Runtime.

https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html

The advantage of implementing a custom runtime is the capability to process multiple events concurrently, but it's unclear if AWS supports this. On the one hand, polling for incoming events is completely within the lambda's control, and an id is used to pair responses with events so they don't necessarily need to be processed in order. But on the other, there are some strange assumptions such as the use of environment variables to pass around tracing headers, which makes me think this is unsupported behavior.

This should be further investigated.

bpholt:
I was wondering if a custom runtime would let us more completely manage resource lifecycles (so that it could run the finalizers when the runtime shuts down, instead of skipping them like it does now). I'm not sure if that's actually supported or not either.

armanbilge:
That's a good idea, that one has bothered me as well. I don't see anything in the custom runtime docs about a shutdown phase, but I did look into this at least a couple times: there is a "Lambda Extensions" API (do you know anything about this?) that does get a shutdown hook:

> Shutdown: This phase is triggered if the Lambda function does not receive any invocations for a period of time. In the Shutdown phase, Lambda shuts down the runtime, alerts the extensions to let them stop cleanly, and then removes the environment. Lambda sends a Shutdown event to each extension, which tells the extension that the environment is about to be shut down.

https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html

It's frustrating if the extension has access to this important lifecycle event, that the lambda itself does not.

armanbilge:
Not official by any means, but I found this:

https://github.com/RhinoSecurityLabs/Cloud-Security-Research/blob/b7d14b7eb51b7dea6ea6a8978682bd9671e0e183/AWS/lambda_ssrf/README.md

> AWS Lambda functions expose an internal API on localhost port 9001 that is meant for custom Lambda runtimes, but it is accessible from within an active invocation of a function. This means that if you get remote code execution or a server side request forgery vulnerability within a Lambda function, you can query that API. The `/2018-06-01/runtime/invocation/next` endpoint will return the event data that was passed into the function for that current execution, where you may be able to find secrets or other useful information to extend your attack into the environment.

[...]

PR #276, opened 2022-09-28
Custom Lambda Runtime

scott-thomson239:
This is a draft PR of my current progress on issue #134 so people can have a look and give feedback/suggestions. I think the main parts that still need work are handling the errors for a possible failure during handler invocation and figuring out how to extract the headers and body from the `Response` gained from the `GET` request sent to the client into a `LambdaRequest`. I also need to test the runtime, which I have left for just now in case something needs changed.


armanbilge (commented):
Thanks for getting this up, this is looking great! Took a pass through with some pointers. Obviously I'm still figuring out some stuff as well :)

Regarding error handling, I guess there are two phases:
1. The setup phase. It looks like we can use the `/runtime/init/error` for this.
2. The per-request phase, for which we can use `/runtime/invocation/AwsRequestId/response`. I think what we want to do is run these on a separate fiber, and wait on the outcome (either success, error, or cancellation), and report it. We can map cancellation to a `CancellationException`.

    Of course, there's still the chance that something goes wrong after initialization but before we obtain the `AwsRequestId` so we cannot use either error reporting mechanism. I think in this instance we have no choice but to crash. AFAICT we cannot proceed to consume the next event since we cannot notify the lambda we are done consuming the current event without the `AwsRequestId`.

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaReservedEnvVars.scala
```diff
@@ -0,0 +1,20 @@
...
+  val AWS_LAMBDA_FUNCTION_VERSION = "AWS_LAMBDA_FUNCTION_VERSION"
+  val AWS_LAMBDA_LOG_GROUP_NAME = "AWS_LAMBDA_LOG_GROUP_NAME"
+  val AWS_LAMBDA_LOG_STREAM_NAME = "AWS_LAMBDA_LOG_STREAM_NAME"
+  val AWS_ACCESS_KEY_ID = "AWS_ACCESS_KEY_ID"
+  val AWS_SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY"
+  val AWS_LAMBDA_RUNTIME_API = "AWS_LAMBDA_RUNTIME_API"
+  val LAMBDA_TASK_ROOT = "LAMBDA_TASK_ROOT"
+  val LAMBDA_RUNTIME_DIR = "LAMBDA_RUNTIME_DIR"
+  val TZ = "TZ"
+}
```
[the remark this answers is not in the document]
scott-thomson239: Hi, I'm not sure I 100% understand this part. Isn't the `Env[F]` from `cats.effect.std` not included in the current version of cats effect? Or do you mean that I should create a custom `Env[F]` for the moment until it becomes available? 

Also, I'm not sure how I would implement the methods within the `LambdaRuntimeEnv[F]` trait. I would have assumed that the implementations of each of the methods would be given here
```scala
object LambdaRuntimeEnv {
	def apply[F[_]](env: Env[F]): LambdaRuntimeEnv[F] = new LambdaRuntimeEnv[F] {
		def lambdaRuntimeDir: F[String] = ???
		// ...
	}
}
```
instead of giving implementations directly within the trait.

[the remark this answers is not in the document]
armanbilge: > Isn't the `Env[F]` from `cats.effect.std` not included in the current version of cats effect?

It is now available in v3.4.0-RC1, you can update to that and use it :)

> Also, I'm not sure how I would implement the methods within the `LambdaRuntimeEnv[F]` trait.

Sorry, I typed that out too hasty :) you are right, the implementations should go in the `apply` method as you wrote. Also I think I meant to write `apply[F[_]](implicit env: Env[F]) ...`

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -14,56 +14,54 @@
...
-        request <- client.get(runtimeUrl)(LambdaRequest.fromResponse) // unsure how to deal with bad response
-        context <- createContext(request)
-        result <- handler(request.body, context)
-        invocationUrl = getInvocationUrl(LambdaReservedEnvVars.AWS_LAMBDA_RUNTIME_API, request.id)
-        _ <- client.successful(POST(result, invocationUrl))
-      } yield ()).foreverM
-    }
-  }.as(()) // how to handle Outcome error and cancellation?
+  def apply[F[_]](client: Client[F])(handler: (Json, Context[F]) => F[Json])(implicit F: Async[F]): F[Unit] = {
+    implicit val lambdaEnv: LambdaRuntimeEnv[F] = LambdaRuntimeEnv(Env.make) // maybe better way
```
armanbilge, line 45: We can just directly request an `implicit env: LambdaRuntimeEnv[F]`

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -14,56 +14,54 @@
...
+    implicit val jsonEncoder: EntityEncoder[F, Json] = jsonEncoderWithPrinter[F](Printer.noSpaces.copy(dropNullValues = true))
+    val http4sClientDsl = new Http4sClientDsl[F] {}
```
armanbilge, line 47: It's a bit superstitious, but would you mind if we avoid using the DSL here and just use the `Request(...)` constructor? Since our use is so small and straightforward, and DSLs always risk to be more unstable than the core APIs (indeed, there was some discussion of moving the routing DSL to a separate repository).

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala
```diff
@@ -0,0 +1,70 @@
...
+  final val AWS_LAMBDA_FUNCTION_MEMORY_SIZE = "AWS_LAMBDA_FUNCTION_MEMORY_SIZE"
+  final val AWS_LAMBDA_FUNCTION_VERSION = "AWS_LAMBDA_FUNCTION_VERSION"
+  final val AWS_LAMBDA_LOG_GROUP_NAME = "AWS_LAMBDA_LOG_GROUP_NAME"
+  final val AWS_LAMBDA_LOG_STREAM_NAME = "AWS_LAMBDA_LOG_STREAM_NAME"
+  final val AWS_ACCESS_KEY_ID = "AWS_ACCESS_KEY_ID"
+  final val AWS_SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY"
+  final val AWS_LAMBDA_RUNTIME_API = "AWS_LAMBDA_RUNTIME_API"
+  final val LAMBDA_TASK_ROOT = "LAMBDA_TASK_ROOT"
+  final val LAMBDA_RUNTIME_DIR = "LAMBDA_RUNTIME_DIR"
+  final val TZ = "TZ"
```
armanbilge, line 39: These can all be `private[this]`

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala
```diff
@@ -0,0 +1,70 @@
...
+
+  def apply[F[_]: Functor](env: Env[F]): LambdaRuntimeEnv[F] = new LambdaRuntimeEnv[F] { // should be error effect?
+    override def handler: F[String] = env.get(HANDLER).map(_.get)
```
armanbilge, line 42: We should avoid `.get`, instead something like this with `MonadThrow[F]`.

```suggestion
    def handler: F[String] = env.get(HANDLER).flatMap(_.liftTo(new NoSuchElementException(HANDLER))
```

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala
```diff
@@ -0,0 +1,70 @@
...
+  final val AWS_LAMBDA_LOG_GROUP_NAME = "AWS_LAMBDA_LOG_GROUP_NAME"
+  final val AWS_LAMBDA_LOG_STREAM_NAME = "AWS_LAMBDA_LOG_STREAM_NAME"
+  final val AWS_ACCESS_KEY_ID = "AWS_ACCESS_KEY_ID"
+  final val AWS_SECRET_ACCESS_KEY = "AWS_SECRET_ACCESS_KEY"
+  final val AWS_LAMBDA_RUNTIME_API = "AWS_LAMBDA_RUNTIME_API"
+  final val LAMBDA_TASK_ROOT = "LAMBDA_TASK_ROOT"
+  final val LAMBDA_RUNTIME_DIR = "LAMBDA_RUNTIME_DIR"
+  final val TZ = "TZ"
+
+  def apply[F[_]: Functor](env: Env[F]): LambdaRuntimeEnv[F] = new LambdaRuntimeEnv[F] { // should be error effect?
```
armanbilge, line 41: Hmm, on second thought, I think we want:

```suggestion
  def apply[F[_]](implicit lre: LambdaRuntimeEnv[F]): LambdaRuntimeEnv[F] = lre

  implicit def forEnv[F[_]: MonadThrow](implicit env: Env[F]): LambdaRuntimeEnv[F] = new LambdaRuntimeEnv[F] {
```

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -14,56 +14,54 @@
...
+    import http4sClientDsl._
+    (for {
+      runtimeApi <- lambdaEnv.lambdaRuntimeApi
+      request <- client.get(getRuntimeUrl(runtimeApi))(LambdaRequest.fromResponse)
+      context <- createContext(request)
+      result <- handler(request.body, context)
+      invocationUrl = getInvocationUrl(LambdaRuntimeEnv.AWS_LAMBDA_RUNTIME_API, request.id)
+      _ <- client.successful(POST(result, invocationUrl))
```
armanbilge, line 55: `successful` returns a `Boolean`, which is currently being ignored. I wonder if `expect[Unit]` will work, so it will raise an exception on failure.

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -73,23 +71,18 @@ object FeralLambdaRuntime {
       request.id,
       logGroupName,
       logStreamName,
-      None, //need
-      None, //need
-      F.delay(FiniteDuration(request.deadlineTimeInMs.toEpochMilli - Instant.now.toEpochMilli, TimeUnit.MILLISECONDS))
+      None,
+      None,
+      F.realTimeInstant.map(curTime => FiniteDuration(request.deadlineTimeInMs.toEpochMilli - curTime.toEpochMilli, TimeUnit.MILLISECONDS))/// how to provide test version?, maybe separate Clock parameter?
```
armanbilge, line 76: Instead of converting to millis and back to duration, you can use `F.realTime` and do that math directly with `FiniteDuration`.

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -73,23 +71,18 @@ object FeralLambdaRuntime {
...
     )
   }
 
-  private def getRuntimeUrl(api: String) = Uri.unsafeFromString(s"http://$api/$LAMBDA_VERSION_DATE/runtime/invocation/next") //need to be unsafe?
+  private def getRuntimeUrl(api: String) = uri"http://$api/$ApiVersion/runtime/invocation/next"
```
armanbilge, line 80: huh, does using `uri"..."` here actually work? That seems like a bug 😅 since it has no way of proving at compile-time that `api` is valid.

build.sbt
```diff
@@ -121,8 +121,9 @@ lazy val lambdaRuntime = crossProject(JSPlatform, JVMPlatform, NativePlatform)
   .crossType(CrossType.Pure)
   .in(file("feral-lambda-runtime"))
   .settings(
-    name := "feral-lambda-kernel",
+    name := "feral-lambda-runtime",
```
armanbilge, line 124: Woops :) thanks! can also rename the file to just `file("lambda-runtime")`

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala
```diff
@@ -23,48 +23,53 @@ trait LambdaRuntimeEnv[F[_]] {
...
 
-    override def lambdaRuntimeDir: F[String] = env.get(LAMBDA_RUNTIME_DIR).map(_.get)
+    override def lambdaTaskRoot: F[String] = getOrThrow(LAMBDA_TASK_ROOT)
 
-    override def timezone: F[String] = env.get(TZ).map(_.get)
+    override def lambdaRuntimeDir: F[String] = getOrThrow(LAMBDA_RUNTIME_DIR)
+
+    override def timezone: F[String] = getOrThrow(TZ)
+
+    private[this] def getOrThrow(envName: String): F[String] =
```
armanbilge, line 72:
```suggestion
    private[this] def getOrRaise(envName: String): F[String] =
```

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRequest.scala
```diff
@@ -1,21 +1,61 @@
...
+    def fromResponse[F[_]](response: Response[F])(implicit F: Concurrent[F]): F[LambdaRequest] = {
+        implicit val jsonDecoder: EntityDecoder[F, Json] = jsonDecoderIncremental
+        for { // TODO use custom header model for aws headers instead
+            id <- F.fromOption(response.headers.get(CIString(requestIdHeader)), new NoSuchElementException(requestIdHeader)).map(_.head.value)
+            invokedFunctionArn <- F.fromOption(response.headers.get(CIString(invokedFunctionArnHeader)), new NoSuchElementException(invokedFunctionArnHeader)).map(_.head.value)
+            deadlineTimeInMs <- F.fromOption(response.headers.get(CIString(deadlineTimeHeader)), new NoSuchElementException(deadlineTimeHeader)).map(_.head.value.toLong)
+            identity <- F.pure(response.headers.get(CIString(cognitoIdentityHeader)).flatMap(_.head.value.asJson.as[CognitoIdentity].toOption))
+            clientContext <- F.pure(response.headers.get(CIString(clientContextHeader)).flatMap(_.head.value.asJson.as[ClientContext].toOption))
+            body <- F.rethrow(jsonDecoder.decode(response, strict = false).value) // basically a reimplementation of response.as[Json] from dsl
```
armanbilge, line 37: `response.as` is not DSL, it's a core method :)

By DSL, I very specifically meant things you need to import from `org.http4s.client.dsl._`

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -18,50 +18,58 @@ package feral.lambda
...
-  }.void
+  }
   
-  private def createContext[F[_]](request: LambdaRequest)(implicit F: Sync[F], lambdaEnv: LambdaRuntimeEnv[F]): F[Context[F]] = for {
-    functionName <- lambdaEnv.lambdaFunctionName
-    functionVersion <- lambdaEnv.lambdaFunctionVersion
-    functionMemorySize <- lambdaEnv.lambdaFunctionMemorySize
-    logGroupName <- lambdaEnv.lambdaLogGroupName
-    logStreamName <- lambdaEnv.lambdaLogStreamName
+  private def createContext[F[_]](request: LambdaRequest)(implicit F: Sync[F], env: LambdaRuntimeEnv[F]): F[Context[F]] = for {
```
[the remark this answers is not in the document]
armanbilge: Eh, since it's a private method you can just re-use the `Temporal` here.

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRequest.scala
```diff
@@ -0,0 +1,23 @@
...
+    id: String,
+    invokedFunctionArn: String,
+    body: Json
+    )
+
+object LambdaRequest {
+    // Still need to decide how to handle failed request or invalid header values
+    def fromResponse[F[_]](response: Response[F])(implicit F: ApplicativeError[F, Throwable]): F[LambdaRequest] = {
+        F.pure(LambdaRequest(???))
+        //Unsure on best way to unpack the response headers into LambdaRequest case class
```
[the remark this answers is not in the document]
scott-thomson239: I had a go at creating header models using a similar way how the headers defined in the `headers` package are defined, but they use functions that are private to the `http4s` package. Is it possible to create header models not using these private functions?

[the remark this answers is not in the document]
armanbilge: Which private methods specifically?

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,104 @@
...
+import org.http4s.implicits.http4sLiteralsSyntax
+
+import scala.concurrent.CancellationException
+
+object FeralLambdaRuntime {
+
+  final val ApiVersion = "2018-06-01"
+
+  // TODO refactor into setup and per-request phases, need to find boundary between two phases
+  def apply[F[_]](client: Client[F])(handler: (Json, Context[F]) => F[Json])(implicit F: Temporal[F], env: LambdaRuntimeEnv[F]): F[Unit] = {
```
armanbilge, line 48: Btw, thanks for raising the issue about loading the handler function. After thinking about it, you are right that we need a small adjustment here actually.

```suggestion
  def apply[F[_]](client: Client[F])(handler: Resource[F, (Json, Context[F]) => F[Json]])(implicit F: Temporal[F], env: LambdaRuntimeEnv[F]): F[Unit] = {
```

Instead of requesting the handler, we should request a `Resource` that is used to acquired the handler. This way, if there are any errors during that process, we can log them correctly. It also leaves the door open for someone to implement dynamic loading, but we shouldn't worry about that here.

feral-lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRequest.scala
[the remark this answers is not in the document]
scott-thomson239: The main one was the `ParseResult.fromParser` method which seems needed to create an instance of `Header`.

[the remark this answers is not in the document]
armanbilge: I think you can just inline that from here, it's pretty straightforward.
https://github.com/http4s/http4s/blob/14afe7a41ee66ef81af4861269b3f574cd9aea8c/core/shared/src/main/scala/org/http4s/MessageFailure.scala#L84-L88

I suspect the `try`/`catch` is no longer necessary, it is an artifact of old code 😂

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
scott-thomson239: I believe the JVM runtime uses reflection based loading of the handler class and the JS runtime uses dynamic loading. Errors can occur during these steps so that's where the initialization error URL is used.

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,94 @@
...
+    )
+  }
+
+  private def getNextInvocationUrl[F[_]: MonadThrow](api: String): F[Uri] = Uri.fromString(s"/$api/$ApiVersion/runtime/invocation/next").liftTo[F]
+
+  private def getInvocationResponseUrl[F[_]: MonadThrow](api: String, id: String): F[Uri] = Uri.fromString(s"/$api/$ApiVersion/runtime/invocation/$id/response").liftTo[F]
+
+  private def getInitErrorUrl[F[_]: MonadThrow](api: String): F[Uri] = Uri.fromString(s"/$api/$ApiVersion/runtime/init/error").liftTo[F]
+
+  private def getInvocationErrorUrl[F[_]: MonadThrow](api: String, requestId: String): F[Uri] = Uri.fromString(s"/$api/$ApiVersion/runtime/invocation/$requestId/error").liftTo[F]
```
armanbilge, line 92: I think we can do `Uri.fromString(s"/$api/$ApiVersion/runtime/")` once, and then safely all the other URIs using that.

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,94 @@
...
+      request <- client.get(nextInvocationUrl)(LambdaRequest.fromResponse[F])
+      invocationErrorUrl <- getInvocationErrorUrl(runtimeApi, request.id) // unsure if errors in these lines can be counted as initialization errors or not, runtime crashes for just now
+      _ <- (for {
+        context <- createContext(request)
+        handlerFiber <- handlerFun(request.body, context).start
+        result <- handlerFiber.join.flatMap {
+          case Succeeded(result: F[Json]) => result
+          case Errored(e: Throwable) => F.raiseError[Json](e)
+          case Canceled() => F.raiseError[Json](new CancellationException)
+        }
```
armanbilge, line 55: I don't think we want to `raiseError`s here, since that will crash the Lambda.

When we join the fiber, if there is an error or cancellation we should send it to the invocation error URL and suppress it.

scott-thomson239: My thinking was that these potential errors would get caught in the `handleErrorWith` method below and instead of crashing the lambda, it would send a invocation error to the invocation error URL. Although I haven't actually tested that this works yet and will need to add another test to see if it can process multiple requests with failures.

armanbilge: Yes, sorry, you are right, I read that much too hastily :) indeed, the `handleErrorWith` will catch these errors. It looks like it will also catch an error POSTing a response.

I guess there could be a distinction between handler errors (since these occur in userland) and errors that occur outside/after the handler, since I assume those would be our fault (or an issue in the runtime)? I actually don't know what the response URL returns if the user handler returns a `Json` that is invalid for this particular Lambda.

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala
```diff
@@ -0,0 +1,64 @@
...
+import org.http4s.Uri
+
+trait LambdaRuntimeEnv[F[_]] {
+  def lambdaFunctionName: F[String]
+  def lambdaFunctionMemorySize: F[Int]
+  def lambdaFunctionVersion: F[String]
+  def lambdaLogGroupName: F[String]
+  def lambdaLogStreamName: F[String]
+  def lambdaRuntimeApi: F[Uri]
+}
```
armanbilge, line 31: Do we know if these values ever change while the process is running? If not maybe we should load them from the `Env` once, and keep them cached.

I cannot imagine why they'd change, but I know Lambda does rely on other envs changing at runtime, so it wouldn't be surprising.

sthomson239: Ah, this might have been a mistake on my end. The [custom lambda runtime docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html) lists reading environment variables as an initialization task while the runtime is currently reading them for each invocation. So no, I don't think they will change so I think it's safe to read them once at the initialization step instead.

scott-thomson239: Ah this might have been my mistake. The [custom lambda runtime docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html) mention that reading environment variables is an initialization task while the custom runtime is currently reading them for every invocation. So no, I don't think they will change so I think it is safe to read them once during initialization.

armanbilge: Thanks! No not at all, I think it was my bad suggestion in https://github.com/typelevel/feral/pull/276#discussion_r982731850 😅 sorry!

So then let's refactor this so they are read once during initialization.

armanbilge: Yes, please go ahead!

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,110 @@
...
+object LambdaRuntime {
+
+  final val ApiVersion = "2018-06-01"
+
+  def apply[F[_]](client: Client[F])(
+      handlerResource: Resource[F, (Json, Context[F]) => F[Json]])(
+      implicit F: Temporal[F],
+      env: LambdaRuntimeEnv[F]): F[Nothing] =
+    handlerResource.attempt.use[Nothing] { handlerOrError =>
+      env.lambdaRuntimeApi.flatMap[Nothing] { api =>
```
scott-thomson239, line 40: Can I ask why `Nothing` is used instead of `Unit` in places such as this? With this code, the compiler complains with 
```
type mismatch;
 found   : Either[Throwable,feral.lambda.runtime.LambdaSettings] => F[Nothing]
 required: Either[Throwable,feral.lambda.runtime.LambdaSettings] => F[B]
```
but runs fine with `Unit`

armanbilge: Good question. Technically we should return `Nothing` here because these methods never return—they start a process that runs forever, and will never produce an output. But perhaps type inference is too uncooperative for this to be practical 🤔

armanbilge: For example, see `Resource#useForever`:
https://github.com/typelevel/cats-effect/blob/52b5a3ba1623ff40fc48318c9a0bfae558f0e5ed/kernel/shared/src/main/scala/cats/effect/kernel/Resource.scala#L217-L224

sideeffffect: I think you should be able to use `Nothing` (and should use it, because, as @armanbilge says, it's the right thing to do).
But you probably need to tweak the variance of `F[_]`. My hunch is that it should be declared as `F[+_]`, so
```scala
private[this] def handleInitError[F[+_]](
```
and similarly for others.
Good luck :crossed_fingers:

armanbilge: Thanks! I also suspect that `F[+_]` would help things. However type signatures of this sort are sort of an anti-pattern in the Typelevel ecosystem (I'm not aware of any other major projects using them) partly because they prevent the use of datatypes that are not declared covariant at the type-level. To be fair, this might not matter in practice as `F[_]` is almost always going to be `IO` ... but I'd rather not constraint the API here when a few explicit type parameters can do the trick, albeit annoying.

sideeffffect: > anti-pattern

Is it really, though? AFAIK, maybe contrary to popular opinion, variance is fairly widely used in the TypeLevel world, isn't it? For example IO in Cats Effect is most definitely declared as `IO[+A]`, as you say, and there are surely many other examples.
But fair enough.

But if you @scott-thomson239 don't want to dive into variance, you shouldn't fix it by using `Unit`. The proper way then is to declare additional type variable which can then by inferred to anything that's necessary, e.g.
```scala
  private[this] def handleInitError[F[_], X](runtimeUri: Uri, client: Client[F], ex: Throwable)(
      implicit F: Temporal[F]): F[X] = {
```

scott-thomson239: Okay no problem, I will try and get `Nothing` working instead of `Unit`. I was just using `Unit` temporarily so that I could fix the failing tests so I will revert that.

Issue #334, opened 2023-03-25
Why not Dispatcher?


mcanlas:
Hello from Scalar 2023 in Warsaw! I saw a presentation by Kamil Kloch about cats-effect and learned about Dispatcher being the go-to mechanism for running IO's in unsafe territory.

I'm in a position where I've home-rolled a solution that is similar to Feral (build a resource, manually initialize, toss away the finalizer bc lifetimes are hard) and in the last mile call `unsafeRunSync`.

The presentation I watched was advocating that there shouldn't be a reason to do that and just use a Dispatcher `.use()` instead "because reasons". Something about the Dispatcher having a better handle on initializing what is required to run IOs?

- A) How is the Dispatcher's run different than just using the unsafe global provided?
- B) Should feral replace its `unsafeRunSync` with a Dispatcher per the advice above? Is there some nuance/extra dimension we are missing?

Edit: I believe Dispatcher thoughts are a continuation on #33

armanbilge:
Thanks for opening the issue! It's a good question, and I think we can revisit this design choice.

> How is the Dispatcher's run different than just using the unsafe global provided?

Here are three advantages of using `Dispatcher#unsafeRun` over `IO#unsafeRun*`:

1. You can have a dispatcher in a generic effect `F[_]` i.e. `Dispatcher[F]`
2. A dispatcher has a scope / lifecycle i.e. it is a `Resource[F, Dispatcher[F]]`. You can be confident that tasks are not being leaked after the dispatcher has been closed.
3. A `Dispatcher` is backed by an already running "warm" fiber. Meanwhile, `IO#unsafeRun*` has to do extra set-up every time it is used, e.g. registering fatal error handlers.

> * Should feral replace its `unsafeRunSync` with a Dispatcher per the advice above?

Yes and no :) point (3) above is a good motivation to create a dispatcher when the lambda starts and use it to handle all incoming requests.

The question is, how will we _get_ that dispatcher? Since the only way to create a `Dispatcher` is inside of an effectful resource, but without the dispatcher we can't run any effects.

So, we will still need `IO#unsafeRun*`. We can use this exactly once to obtain a dispatcher during the setup process, that we can then use for all subsequent requests.

Does that answer your question? Thanks, and PRs welcome for this :)

mcanlas:
So are you suggesting that with both Feral and my work, we should bootstrap a Dispatcher like a resource for the lifetime of the lambda, and then continually use that Dispatcher for the times when we need to be unsafe?

[...]

PR #335, opened 2023-03-27
Use `Dispatcher` for invoking lambda handler

mcanlas:
Per #334

### DIFF
--- a/core/src/main/scala/feral/IOSetup.scala
+++ b/core/src/main/scala/feral/IOSetup.scala
@@ -16,10 +16,13 @@
 
 package feral
 
+import scala.concurrent.Future
+
 import cats.effect.unsafe.IORuntime
 import cats.effect.kernel.Resource
 import cats.effect.IO
-import cats.effect.kernel.Deferred
+import cats.effect.std.Dispatcher
+import cats.syntax.all._
 
 private[feral] trait IOSetup {
 
@@ -28,17 +31,11 @@ private[feral] trait IOSetup {
   protected type Setup
   protected def setup: Resource[IO, Setup] = Resource.pure(null.asInstanceOf[Setup])
 
-  private[feral] final lazy val setupMemo: IO[Setup] = {
-    val deferred = Deferred.unsafe[IO, Either[Throwable, Setup]]
-    setup
-      .attempt
+  private[feral] final lazy val setupMemo: Future[(Dispatcher[IO], Setup)] =
+    (Dispatcher.parallel[IO](await = false), setup)
+      .tupled
       .allocated
-      .flatTap {
-        case (setup, _) =>
-          deferred.complete(setup)
-      }
-      .unsafeRunAndForget()(runtime)
-    deferred.get.rethrow
-  }
+      .map(_._1) // drop unused finalizer
+      .unsafeToFuture()(runtime)
 
 }
--- a/lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -34,10 +34,15 @@ private[lambda] trait IOLambdaPlatform[Event, Result] {
   private lazy val handlerFn
       : js.Function2[js.Any, facade.Context, js.Promise[js.Any | Unit]] = {
     (event: js.Any, context: facade.Context) =>
-      (for {
-        lambda <- setupMemo
-        event <- IO.fromEither(decodeJs[Event](event))
-        result <- lambda(event, Context.fromJS(context))
-      } yield result.map(_.asJsAny).orUndefined).unsafeToPromise()(runtime)
+      setupMemo.toJSPromise(scala.concurrent.ExecutionContext.parasitic).`then`[js.Any] {
+        case (dispatcher, lambda) =>
+          val io =
+            for {
+              event <- IO.fromEither(decodeJs[Event](event))
+              result <- lambda(event, Context.fromJS(context))
+            } yield result.map(_.asJsAny).orUndefined
+
+          dispatcher.unsafeToPromise(io)
+      }
   }
 }
--- a/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -16,6 +16,9 @@
 
 package feral.lambda
 
+import scala.concurrent.Await
+import scala.concurrent.duration.Duration
+
 import cats.data.OptionT
 import cats.effect.IO
 import cats.effect.kernel.Resource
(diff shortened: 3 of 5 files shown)


core/src/main/scala/feral/IOSetup.scala
```diff
@@ -20,24 +20,36 @@ import cats.effect.unsafe.IORuntime
 import cats.effect.kernel.Resource
 import cats.effect.IO
 import cats.effect.kernel.Deferred
+import cats.effect.std.Dispatcher
 
 private[feral] trait IOSetup {
+  private def runtime =
+    IORuntime.global
 
-  protected def runtime: IORuntime = IORuntime.global
```
armanbilge, line 26: We should keep the `runtime` as an overridable `protected def`, in case the user wants to change it.

core/src/main/scala/feral/IOSetup.scala
```diff
@@ -20,24 +20,36 @@ import cats.effect.unsafe.IORuntime
...
+  protected def dispatcher: Dispatcher[IO] =
```
armanbilge, line 29: Meanwhile, this should probably be `final`. I am not sure in what case it should be overriden.

core/src/main/scala/feral/IOSetup.scala
```diff
@@ -20,24 +20,28 @@ import cats.effect.unsafe.IORuntime
...
+    Dispatcher.sequential[IO](await = true).allocated.map(_._1).unsafeRunSync()(runtime)
```
armanbilge, line 30: `await = true` doesn't really matter, since we are discarding the finalizer anyway.

Instead of `sequential` I would lean for `parallel`. It slightly increases the overhead per invocation, but it avoids the situation where one task prevents another task from running.

Also we can't use `unsafeRunSync()` anywhere in this code, because it must cross-compile for JS as well.

core/src/main/scala/feral/IOSetup.scala
```diff
@@ -20,24 +20,28 @@ import cats.effect.unsafe.IORuntime
...
 
   protected type Setup
   protected def setup: Resource[IO, Setup] = Resource.pure(null.asInstanceOf[Setup])
 
   private[feral] final lazy val setupMemo: IO[Setup] = {
     val deferred = Deferred.unsafe[IO, Either[Throwable, Setup]]
```
armanbilge, line 36: Since this is a one time setup, there's not much reason to run it on the `Dispatcher`. In fact, we could setup the `Dispatcher` and the memo in the same unsafe call.

armanbilge:
Thanks for the PR! If I may suggest an alternative approach:
https://github.com/typelevel/feral/blob/a99c5d15490ed08a0edb55322c0b4f39f3cdfcb1/core/src/main/scala/feral/IOSetup.scala#L31

What if we change that to return a `Future[(Dispatcher[IO], Setup)]` ?

mcanlas:
I follow the top comments but less the bottom ones. How do you end up allocating the dispatcher if we can't `unsafeRunSync`? Also where does the `Future[_]` come from now?

armanbilge:
`unsafeRunToFuture()` :)

mcanlas:
I don't think I totally follow. Wouldn't that have knock on effects to where `setupMemo` is currently consumed? What do they change to?

mcanlas:
Still not totally understanding what the `Await` part should look like, as well as the combining with the promises 😩

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -34,10 +36,24 @@ private[lambda] trait IOLambdaPlatform[Event, Result] {
...
     (event: js.Any, context: facade.Context) =>
-      (for {
-        lambda <- setupMemo
-        event <- IO.fromEither(decodeJs[Event](event))
-        result <- lambda(event, Context.fromJS(context))
-      } yield result.map(_.asJsAny).orUndefined).unsafeToPromise()(runtime)
+      {
+        // how to extract?
+        // need to combinate with below?
+        setupMemo.toJSPromise(scala.concurrent.ExecutionContext.global)
```
armanbilge, line 42: So the same way you can combine `Future` with `flatMap`, you can combine `js.Promise` with `then`.

Also, you can use `runtime.compute` instead of `ExecutionContext.global`.

lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -34,29 +37,33 @@ private[lambda] abstract class IOLambdaPlatform[Event, Result]
...
-            .head
-            .compile
-            .lastOrError
-          context <- IO(Context.fromJava[IO](context))
-          _ <- OptionT(lambda(event, context)).foreachF { result =>
-            Resource.fromAutoCloseable(IO(new OutputStreamWriter(output))).use { writer =>
-              IO.blocking(Printer.noSpaces.unsafePrintToAppendable(result.asJson, writer))
+      context: lambdaRuntime.Context): Unit = {
+    val (dispatcher, lambda) =
+      Await.result(setupMemo, Duration.Inf)
```
armanbilge, line 42: This is right! :)

mcanlas:
The given error I'm not even sure how to fix:

```
[error] /home/runner/work/feral/feral/lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala:47:37: type mismatch;
[error]  found   : scala.scalajs.js.Promise[scala.scalajs.js.UndefOr[scala.scalajs.js.Any]]
[error]     (which expands to)  scala.scalajs.js.Promise[scala.scalajs.js.Any | Unit]
[error]  required: ? | scala.scalajs.js.Thenable[?]
```

Is `then` the right combinator and does it just take a function-like body? Is a `Promise` not also a `Thenable`? What is it complaining about?

lambda/js/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -34,10 +36,15 @@ private[lambda] trait IOLambdaPlatform[Event, Result] {
   private lazy val handlerFn
       : js.Function2[js.Any, facade.Context, js.Promise[js.Any | Unit]] = {
     (event: js.Any, context: facade.Context) =>
-      (for {
-        lambda <- setupMemo
-        event <- IO.fromEither(decodeJs[Event](event))
-        result <- lambda(event, Context.fromJS(context))
-      } yield result.map(_.asJsAny).orUndefined).unsafeToPromise()(runtime)
+      setupMemo.toJSPromise(scala.concurrent.ExecutionContext.parasitic).`then` {
```
armanbilge, line 39: Try this, I think it's just type inference problems :)

```suggestion
      setupMemo.toJSPromise(scala.concurrent.ExecutionContext.parasitic).`then`[js.Any] {
```

armanbilge:
For this one it should be fine to use `IO.fromFuture`.

https://github.com/typelevel/feral/blob/a99c5d15490ed08a0edb55322c0b4f39f3cdfcb1/lambda/shared/src/test/scala/feral/lambda/IOLambdaSuite.scala#L37-L37

armanbilge:
I pushed a commit adding a MiMa binary-compatibility filter for your changes.

armanbilge (approved):
Great job identifying this optimization opportunity, I had totally missed it! Thanks for all your work on this complex, cross-platform refactor.

mcanlas:
Credit should go to Kamil Kloch and his presentation at Scalar, and by extension the folks in the Discord that helped advocate for Dispatcher as The Way

PR #354, opened 2023-06-02
Add S3Event

shagoon:
This PR adds case classes and json-decoders for `S3Event`. This allows S3 to trigger the lambda function as described [here](https://docs.aws.amazon.com/lambda/latest/dg/with-s3-example.html).

### DIFF
--- a/lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
@@ -1,0 +1,145 @@
+/*
+ * Copyright 2021 Typelevel
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.lambda.events
+
+import io.circe.Decoder
+
+import java.time.Instant
+
+// https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/trigger/s3.d.ts
+
+final case class S3Event(records: List[S3EventRecord])
+
+object S3Event {
+  implicit val decoder: Decoder[S3Event] = Decoder.forProduct1("Records")(S3Event.apply)
+}
+
+final case class S3EventRecord(
+    eventVersion: String,
+    eventSource: String,
+    awsRegion: String,
+    eventTime: Instant,
+    eventName: String,
+    userIdentity: S3UserIdentity,
+    requestParameters: S3RequestParameters,
+    responseElements: S3ResponseElements,
+    s3: S3,
+    glacierEventData: Option[S3EventRecordGlacierEventData])
+
+object S3EventRecord {
+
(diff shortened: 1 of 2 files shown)


armanbilge (commented):
Thanks!

lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
```diff
@@ -0,0 +1,145 @@
...
+    eventName: String,
+    userIdentity: S3UserIdentity,
+    requestParameters: S3RequestParameters,
+    responseElements: S3ResponseElements,
+    s3: S3,
+    glacierEventData: Option[S3EventRecordGlacierEventData])
+
+object S3EventRecord {
+
+  implicit private val instantDecoder: Decoder[Instant] = Decoder[String].map(Instant.parse(_))
```
armanbilge, line 43: Can we use this one?
https://github.com/circe/circe/blob/172c9a4d9e14270bab59700d9ffe439462c23db7/modules/core/shared/src/main/scala/io/circe/Decoder.scala#L1241-L1244

shagoon: Probably, looks quite similar. I'll try, the test covers that.

lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
```diff
@@ -0,0 +1,145 @@
...
+
+object S3EventRecordGlacierEventData {
+
+  implicit val decoder: Decoder[S3EventRecordGlacierEventData] =
+    Decoder.forProduct1("restoreEventData")(S3EventRecordGlacierEventData.apply)
+
+}
+
+final case class S3EventRecordGlacierRestoreEventData(
+    lifecycleRestorationExpiryTime: String,
```
armanbilge, line 135: Should this also be an `Instant`?

lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
shagoon: Yes, that works. Though, I had to keep the implicit val to prevent `feral.lambda.events.instantDecoder` from being used.

armanbilge: Oh I see, there was a conflict with this one 🤔 
https://github.com/typelevel/feral/blob/cbd83badfe6f5af45bc102abe12bc704cb02563a/lambda/shared/src/main/scala/feral/lambda/events/package.scala#L26-L27

Maybe we should make that one not implicit 🤔 anyway, that can be a follow-up PR.

lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
shagoon: Yes, according to [Event message structure](https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html) it's a time in ISO-8601 format as well.

lambda/shared/src/main/scala/feral/lambda/events/S3Event.scala
shagoon: #356 could be that follow-up (or preparation) PR for this one.

armanbilge (approved):
Thank you for your contribution! I will release this shortly.

ecastilla95:
Hello, I am trying to use this event in my Scala.js project but it is telling me that the import doesn't exist.

I am comparing this PR to https://github.com/typelevel/feral/pull/231/ and I wonder if this Event is missing the changes in
lambda/shared/src/main/scala-2/feral/lambda/package.scala
(and maybe also scala-3/feral/lambda/envs.scala)
because my Scala.js version is using Scala 2.12

I am not sure though, because this is my first time encountering this kind of issue.

PR #356, opened 2023-06-05
Make `Decoder[Instant]` explicit

shagoon:
Preparation for #354 

If this gets merged before #354, #354 should be updated to remove the explicit references to `io.circe.Decoder.decodeInstant`.

If #354 gets merged before this one, this one should be updated to remove those references to `io.circe.Decoder.decodeInstant` in `S3EventRecord` and `S3EventRecordGlacierRestoreEventData`.

### DIFF
--- a/lambda/shared/src/test/scala/feral/lambda/events/InstantDecoderSuite.scala
+++ b/lambda/shared/src/test/scala/feral/lambda/events/InstantDecoderSuite.scala
@@ -16,7 +16,7 @@
 
 package feral.lambda.events
 
-import io.circe.Json
+import io.circe.{Decoder, Json}
 import munit.ScalaCheckSuite
 import org.scalacheck.Arbitrary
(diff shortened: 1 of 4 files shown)


lambda/shared/src/test/scala/feral/lambda/events/InstantDecoderSuite.scala
```diff
@@ -26,6 +26,8 @@ import java.time.Instant
 
 class InstantDecoderSuite extends ScalaCheckSuite {
 
+  implicit private val instantDecoder: Decoder[Instant] = feral.lambda.events.instantDecoder
```
armanbilge, line 29:
```suggestion
  implicit private def instantDecoder: Decoder[Instant] = feral.lambda.events.instantDecoder
```

armanbilge (approved):
Thanks!

Issue #48, opened 2021-11-12
Create models for the myriad of AWS Lambda event types


armanbilge:
For feral-lambda to be useful for the "zoo" of Lambda types available on AWS, we need to provide models for the various events (aka "triggers") and responses used by each of these lambdas. These will live in the `feral-lambda-events` module.

We are currently sourcing them from the [Java](https://github.com/aws/aws-lambda-java-libs/tree/master/aws-lambda-java-events) and [JavaScript/TypeScript](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/aws-lambda/trigger) implementations.

The TypeScript definitions are _especially_ important, because they are the definitive representation of these events as "plain old JavaScript objects" which is isomorphic to their JSON representation. Therefore, our models should match them exactly in name, structure, etc. in order for parsing to work correctly.

For an example of this, see the existing events:
https://github.com/typelevel/feral/blob/main/lambda-events/src/main/scala/feral/lambda/events/

fredshonorio:
@armanbilge we talked on discord, for the purposes of project management I'll be looking at the SQS event first.

armanbilge:
@bpholt I've again arrived at the opinion that the `CloudFormationCustomResourceEvent` should probably join the rest of the events in this module 😅 these are my assumptions:

1. Not every event requires its own special lambda (like we have for http4s or CloudFormation). So this justifies the existence of the `feral-lambda-events` module, to host these events.
2. For consistency/least surprise, if some events are in `feral-lambda-events`, they should all be there :)

I could be wrong about (1). When we figure out how to deploy these things, maybe each type of Lambda will have to be treated differently anyway? I hope not.

ecastilla95:
I might have found a bug with the S3Event type
https://github.com/typelevel/feral/pull/354#issuecomment-2137117702

VAIBHAVSING:
@armanbilge  scala implementation to triggering various aws event like triger aws lambda function, triger aws s3  and dynodb.

Yummy-Yums:
@armanbilge I realized others are working on other event types, can we create a sub issue for each event type and assign each, so others can know someone is working on  problem? I almost picked another event to work but i just found out  others are active on it

PR #409, opened 2023-09-20
Add `APIGatewayProxyRequestEvent`

andre-bisa:
This PR allows to create Lambda responding to an API Gateway using the Proxy integration. (linked to #48)


andre-bisa:
Hi @armanbilge,
It's mainly because the HTTP API cannot be set as private. We have a scenario where we use a private REST API with Proxy integration which still uses the old implementation (AFAIK V2 is only supported in the HTTP API).
I think this could be useful to other people as well tho.

armanbilge (approved):
Thanks for explaining! I guess for completeness we might consider adding an http4s integration for this event too, but that doesn't need to be this PR.

PR #424, opened 2023-11-18
Use `ByteVector` for B and BS in DynamoDbStreamEvent

filosganga:
This changes the decoding and representation of  the DynamoDb AttributeValue B and BS. Also this solves a big where the BS DynamoDB AttributeValue BS type was modelled as a single String that would fail the decoding.

B is a binary value Base64 encoded on the wire, while BS is an ordered set of binary base64 encoded value. 

Given that feral depends already on scodec, it makes sense to use BinaryVector to model the binary value.

### DIFF
--- a/lambda/shared/src/test/scala/feral/lambda/events/DynamoDbStreamEventSuite.scala
+++ b/lambda/shared/src/test/scala/feral/lambda/events/DynamoDbStreamEventSuite.scala
@@ -16,13 +16,48 @@
 
 package feral.lambda.events
 
+import cats.syntax.option._
 import io.circe.literal._
 import munit.FunSuite
+import java.nio.charset.StandardCharsets
+import scodec.bits.ByteVector
 
 class DynamoDbStreamEventSuite extends FunSuite {
 
   test("decoder") {
     event.as[DynamoDbStreamEvent].toTry.get
+  }
+
+  test("AttributeValue should decod a B") {
+
+    val vector = ByteVector.view("foo".getBytes(StandardCharsets.UTF_8))
+
+    val source = json"""
+      {
+        "B": ${vector.toBase64}
+      }
+    """
+
+    val decoded = source.as[AttributeValue].toOption.flatMap(_.b)
+
+    assertEquals(decoded, vector.some)
+  }
+
+  test("AttributeValue should decod a BS") {
+
+    val vector = ByteVector.view("foo".getBytes(StandardCharsets.UTF_8))
+
+    val source = json"""
+      {
+        "BS": [
+          ${vector.toBase64}
+        ]
+      }
+    """
+
+    val decoded = source.as[AttributeValue].toOption.flatMap(_.bs)
+
+    assertEquals(decoded, List(vector).some)
   }
 
   def event = json"""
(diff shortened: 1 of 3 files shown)


lambda/shared/src/test/scala/feral/lambda/events/DynamoDbStreamEventSuite.scala
```diff
@@ -16,15 +16,50 @@
...
+import java.nio.charset.StandardCharsets
+import scodec.bits.ByteVector
 
 class DynamoDbStreamEventSuite extends FunSuite {
 
   test("decoder") {
     event.as[DynamoDbStreamEvent].toTry.get
   }
 
+  test("AttributeValue should decod a B") {
```
zetashift, line 31:
```suggestion
  test("AttributeValue should decode a B") {
```

lambda/shared/src/test/scala/feral/lambda/events/DynamoDbStreamEventSuite.scala
```diff
@@ -16,15 +16,50 @@
...
+        "B": ${vector.toBase64}
+      }
+    """
+
+    val decoded = source.as[AttributeValue].toOption.flatMap(_.b)
+
+    assertEquals(decoded, vector.some)
+  }
+
+  test("AttributeValue should decod a BS") {
```
zetashift, line 46:
```suggestion
  test("AttributeValue should decode a BS") {
```

armanbilge:
We need to bump the base version to 0.3 because this is a breaking change.

https://github.com/typelevel/feral/blob/424590163f05fd4d06692c0880815eff387708db/build.sbt#L21

armanbilge:
> Also this solves a big where the BS DynamoDB AttributeValue BS type was modelled as a single String that would fail the decoding.

Sorry, I couldn't quite understand. Would you mind sharing an example? Did we add a test for this case?

filosganga:
> > Also this solves a big where the BS DynamoDB AttributeValue BS type was modelled as a single String that would fail the decoding.
> 
> Sorry, I couldn't quite understand. Would you mind sharing an example? Did we add a test for this case?

Sure, 

B on the wire is a Base64 encoded binary, and at this moment it is modelled as a string. Not ergonomic but it works. 

```json
{
  "B": "Base64..."
}
```

On the other hands BS on the wire is an array of Base64 encoded binary strings:

```json
{
  "BS": [
    "Base64...",
    "Base64..."
  ]
}
```

But, right now, on feral, it is decoded as a single String, and so the circe decoding fails.

Does it make sense?

armanbilge:
Oh, I totally missed that! That is indeed a bug, not related to `String` vs `Bytevector` but because we forgot `List` 🤦 thanks!

lambda/shared/src/test/scala/feral/lambda/events/DynamoDbStreamEventSuite.scala
```diff
@@ -16,15 +16,50 @@
...
+        "B": ${vector.toBase64}
+      }
+    """
+
+    val decoded = source.as[AttributeValue].toOption.flatMap(_.b)
+
+    assertEquals(decoded, vector.some)
+  }
+
+  test("AttributeValue should decode a BS") {
```
armanbilge, line 46: And I see we have a test for it now 👍

filosganga:
> Oh, I totally missed that! That is indeed a bug, not related to `String` vs `Bytevector` but because we forgot `List` 🤦 thanks!

Yeah sorry I have conflated two things given that it was a breaking change anyway

armanbilge (approved):
Thank you so much!

PR #425, opened 2023-11-20
Re-encode `Context` and events for better compatibility and mockability

armanbilge:
This uses the pattern of `sealed abstract class`es with `apply` methods as the public interface and `private final case class`es as the actual implementations.

While passing through I did a few other minor tidy-ups.


armanbilge:
One thing that's missing from this PR are `withX` builder-style methods. I'm not sure how important these are, and we can always add them as-needed without breaking compatibly.

PR #427, opened 2023-11-21
RFC: rename `LambdaEnv` -> `Invocation`

armanbilge:
I think "invocation" is a more suitable:
- the `event` and `context` are relevant only during a specific invocation of the lambda function
- "env" is reminiscent of environment variables that have a different lifecycle. Also the concept became [overloaded in our custom lambda runtime](https://github.com/typelevel/feral/blob/89d1c69f9bfd8b49ab6e6ebe298dfe1822f20376/lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntimeEnv.scala#L24-L31) which is less than ideal.

I left deprecated aliases in place. Should be a pretty easy find-replace for downstreams.


fredshonorio:
Having superficial experience with aws lambda (before feral existed) I like the name better.

bpholt:
Seems reasonable. Could we provide a Scalafix for the rewrite?

armanbilge:
> Could we provide a Scalafix for the rewrite?

Let me look into that :)

armanbilge:
Will do a follow-up PR with scalafixes for this and some other changes. Thanks everyone!

PR #436, opened 2023-12-01
Deprecate KinesisStreamEvent

etspaceman:


armanbilge (approved):
thanks for adopting this event in kinesis4cats :)

PR #437, opened 2023-12-01
Optimize `IOLambda#handleRequest` on JVM

armanbilge:
If we assume* that the `InputStream` and `OutputStream` passed to the handler are a `ByteArrayInputStream` and a `ByteArrayOutputStream` then we can take advantage of the fact that they are non-blocking and actually non-streaming. ~~(In the unlikely case that this assumption is wrong, blocking on the compute before/after the handler seems pretty harmless anyway, since there is no concurrency).~~

This also allows us to drop the circe-fs2 dependency which is [deprecated](https://github.com/circe/circe-fs2#deprecated).

\* FTR:
- https://github.com/aws/aws-lambda-java-libs/blob/3d8dfb66f3a852bf69618bd5f66222692f1f5a49/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/runtimeapi/InvocationRequest.java#L94-L96
- https://github.com/aws/aws-lambda-java-libs/blob/3d8dfb66f3a852bf69618bd5f66222692f1f5a49/aws-lambda-java-runtime-interface-client/src/main/java/com/amazonaws/services/lambda/runtime/api/client/AWSLambda.java#L243-L245

### DIFF
--- a/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -17,7 +17,6 @@
 package feral.lambda
 
 import cats.effect.IO
-import cats.syntax.all._
 import com.amazonaws.services.lambda.{runtime => lambdaRuntime}
 import io.circe.Printer
 import io.circe.jawn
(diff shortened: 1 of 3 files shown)


lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -16,53 +16,41 @@
...
+    val context = Context.fromJava[IO](runtimeContext)
+    dispatcher
+      .unsafeRunTimed(
+        lambda(event, context),
+        runtimeContext.getRemainingTimeInMillis().millis
+      )
+      .foreach { result =>
+        val writer = new OutputStreamWriter(output)
+        Printer.noSpaces.unsafePrintToAppendable(result.asJson, writer)
+        writer.flush()
```
armanbilge, line 52: Ok, so it's a bit unidiomatic to move this and the event decoding out of `IO`. But here's my thinking ...

> In the unlikely case that this assumption is wrong, blocking on the compute before/after the handler seems pretty harmless anyway, since there is no concurrency.

We already know we have to block the lambda "main" thread while we run `IO` on the CE runtime. So by moving all interaction with `input`/`output` outside of `IO` to this thread, if they _are_ blocking (they're not) then at least we're doing it on this already-blocked thread and not clogging up a compute thread.

lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -16,53 +16,41 @@
...
-        }
-        .onFinalize(IO.blocking(input.close()) &> IO.blocking(output.close()))
-        .use_
-    }
+    val event = jawn.decodeChannel[Event](Channels.newChannel(input)).fold(throw _, identity(_))
+    val context = Context.fromJava[IO](runtimeContext)
+    dispatcher
+      .unsafeRunTimed(
+        lambda(event, context),
+        runtimeContext.getRemainingTimeInMillis().millis
```
AlexITC, line 47: This looks suspicious, by this time, AWS could have already killed the lamda, giving us no chance to clear any resources.

armanbilge: > but I wouldn't hardcode this for everyone,an idea is to make this configurable.

exactly! this is why the `.timeout` API exists on `IO` :) anyone can use it however they want.

PR #446, opened 2024-01-11
Remove core module

armanbilge:
This module was never public-facing: it only contained the private `IOSetup` that contained logic to unsafely initialize the `Dispatcher` and acquire `Resource`s. I'm no longer convinced that sharing this logic is worthwhile since we can do better things in platform-specific code.

- On the JVM we can do this initialization by blocking during construction of the handler class. This is better, because it means the AWS Lambda runtime will properly recognize the initialization period and not begin sending events until it is complete.

- On JS I'm hoping we can do something similar with a top-level `await`, see https://github.com/typelevel/feral/issues/246#issuecomment-1836993659.


bpholt:
Makes sense to me!

PR #448, opened 2024-01-15
Implement workaround for initialization order errors in JVM lambdas

armanbilge:
Fixes a regression [reported on Discord](https://discord.com/channels/632277896739946517/918373380003082250/1196217785781592105).

Previously, `IOLambda` was not sensitive to whether initializing the `handler` involved `val`s. This is because the definition of `handler` was used lazily on the first invocation of the lambda, not during its construction.

However, after the changes in https://github.com/typelevel/feral/pull/446, we now use the definition of `handler` eagerly during construction. A `val` in a superclass (i.e. `IOLambda`) is always initialized before `val`s in the subclass (i.e. user-defined `FooLambda`). This means that all `val`s in the subclass are uninitialized (i.e. `null`) when we initialize `val handle` in `IOLambda`.

The proper remedy for this situation is that users should always use `def` instead of `val` in their `IOLambda`s. However, because this is a highly surprising regression wrt the previous behavior, we now make a best-effort to detect this situation, log an actionable warning, and fallback to the old semantic where we initialize during the first invocation of the lambda.

```
There was an error initializing `FooLambda` during startup.
Falling back to initialize-during-first-invocation strategy.
To fix, try replacing any `val`s in `FooLambda` with `def`s.
```

I've published this PR as `0.3-97de28d-SNAPSHOT`.

### DIFF
--- a/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
+++ b/lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
@@ -43,8 +43,9 @@ private[lambda] abstract class IOLambdaPlatform[Event, Result]
         try this.handler
         catch { case ex if NonFatal(ex) => null }
 
-      if (h ne null) { h.map(IO.pure(_)) }
-      else {
+      if (h ne null) {
+        h.map(IO.pure(_))
+      } else {
         val lambdaName = getClass().getSimpleName()
         val msg =
           s"""|There was an error initializing `$lambdaName` during startup.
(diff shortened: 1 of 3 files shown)


lambda/jvm/src/main/scala/feral/lambda/IOLambdaPlatform.scala
```diff
@@ -29,11 +32,33 @@ import java.io.OutputStream
...
+        val msg =
+          s"""|There was an error initializing `$lambdaName` during startup.
+              |Falling back to initialize-during-first-invocation strategy.
+              |To fix, try replacing any `val`s in `$lambdaName` with `def`s.""".stripMargin
+        System.err.println(msg)
+
+        Async[Resource[IO, *]].defer(this.handler).memoize.map {
+          case Resource.Eval(ioa) => ioa
+          case _ => throw new AssertionError
+        }
```
armanbilge, line 58: There will be an official way to do this soon.
- https://github.com/typelevel/cats-effect/pull/3890

Issue #340, opened 2023-04-16
Integrate feral-lambda with otel4s


armanbilge:
Building on https://github.com/typelevel/feral/issues/339, we can then integrate otel4s directly into `IOLambda`, following these additional semantic conventions:

- https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/instrumentation/aws-lambda/

alexcardell:
@armanbilge I am happy to have a go at this, given the otel4s SDK looks to land soon 

What are your thoughts on pulling out a natchez module, and supporting both for the near term?

armanbilge:
> What are your thoughts on pulling out a natchez module, and supporting both for the near term?

This seems like a smart idea, nice! I see you already got a PR up. @bpholt wdyt?

bpholt:
Sounds like a great idea to me! Thanks for submitting the PR @alexcardell.

alexcardell:
The module on its own doesn't seem valuable and an unnecessary breaking change, I've closed it for now and included it in #459 

Up to you guys if you want to reopen and include it in the next breaking, then the otel4s module could be a patch

Issue #339, opened 2023-04-16
Add feral-otel4s integration


armanbilge:
OTel defines some semantic conventions for FaaS "function-as-a-service". A feral-otel4s integration module might be a good place for helpers to live.

- https://opentelemetry.io/docs/reference/specification/metrics/semantic_conventions/faas-metrics/
- https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/faas/

PR #456, opened 2024-02-22
Extract lambda-natchez module

alexcardell:
This is to allow consumers the choice of natchez or otel4s, related to supporting #339 #340 

No breaking changes except requiring a `feral.lambda.natchez._` import


alexcardell:
Superceded by #459

PR #466, opened 2024-03-11
Fix query parameters avoiding duplicates

andre-bisa:
queryStringParameters and multiValueQueryStringParameters were merged together, but multiValueQueryStringParameters already contains all queryStringParameters

### DIFF
--- a/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
+++ b/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
@@ -63,6 +63,7 @@ object ApiGatewayProxyHandler {
       event: ApiGatewayProxyEvent): F[Request[F]] = {
     val queryString: String = getMultiValueQueryStringParameters(
       event.multiValueQueryStringParameters)
+
     val uriString: String = event.path + (if (queryString.nonEmpty) s"?$queryString" else "")
 
     for {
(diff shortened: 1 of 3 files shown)


armanbilge:
Woops, good catch! I think we have the same problem with headers, would you mind fixing that as well?

andre-bisa:
@armanbilge I've fixed headers as well, considering only multivalues. I've also included tests for both (query params and headers)

lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
```diff
@@ -114,6 +95,15 @@ object ApiGatewayProxyHandler {
...
     }
 
+  private def getMultiValueHeaders(
+      multiValueHeaders: Option[Map[CIString, List[String]]]): List[Header.Raw] = {
+    for {
+      headerMap <- multiValueHeaders.toList
+      (key, values) <- headerMap
+      headerRaw <- values.map(value => Header.Raw(key, value))
+    } yield headerRaw
+  }
```
armanbilge, line 105: This implementation uses a lot of intermediate allocations, would you mind re-using the original `ListBuilder`-based implementation?

lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandler.scala
```diff
@@ -61,29 +62,15 @@ object ApiGatewayProxyHandler {
...
-        event.headers.foreach { h => h.foreachEntry(builder += Header.Raw(_, _)) }
-        event.multiValueHeaders.foreach { hMap =>
-          hMap.foreach {
-            case (key, values) =>
-              if (!event.headers.exists(_.contains(key))) {
-                values.foreach(value => builder += Header.Raw(key, value))
-              }
-          }
-        }
-        Headers(builder.result())
```
armanbilge, line 85: This one.

andre-bisa: Done

armanbilge (approved):
thanks for your work!

PR #467, opened 2024-03-15
Add headers to proxy integration

merens:
I had trouble rendering a page because content-type is not correctly passed to API gateway when using REST API.
This problem is not present when using HTTP integration and I noticed that _headers_ is missing.
[Documentation on integration response](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-integration-settings-integration-response.html).

I used the same implentation as [ApiGatewayProxyStructuredResultV2](https://github.com/typelevel/feral/blob/main/lambda-http4s/src/main/scala/feral/lambda/http4s/ApiGatewayProxyHandlerV2.scala#L58).

### DIFF
--- a/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayProxyResult.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayProxyResult.scala
@@ -35,6 +35,10 @@ object ApiGatewayProxyResult {
       isBase64Encoded: Boolean): ApiGatewayProxyResult =
     new Impl(statusCode, headers, body, isBase64Encoded)
 
+  @deprecated("This constructor does not have headers", "0.3.0")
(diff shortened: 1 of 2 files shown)


andre-bisa:
+1

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayProxyResult.scala
```diff
@@ -17,26 +17,35 @@
...
   def statusCode: Int
+  def headers: Map[CIString, String]
   def body: String
   def isBase64Encoded: Boolean
 }
 
 object ApiGatewayProxyResult {
 
-  def apply(statusCode: Int, body: String, isBase64Encoded: Boolean): ApiGatewayProxyResult =
-    new Impl(statusCode, body, isBase64Encoded)
```
armanbilge, line 30: Woops, good catch! Can you keep a `@deprecated` version of the old constructor too, for backwards-compatibility? Thank you!

PR #476, opened 2024-04-17
Add `ApiGatewayV2WebSocketEvent`

Chingles2404:


lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,304 @@
...
+    stageVariables: Map[String, String],
+    requestContext: WebSocketRequestContext,
+    body: String,
+    isBase64Encoded: Boolean
+  ) extends APIGatewayV2WebSocketEvent {
+    override def productPrefix = "APIGatewayV2WebSocketEvent"
+  }
+}
+
+sealed abstract class WebSocketRequestIdentity {
```
armanbilge, line 100: I didn't look too carefully. How similar is this one to the existing `RequestIdentity`? If they are the same maybe we can share the classes instead of implementing it twice.

armanbilge (commented):
Thanks for working on this! Sorry for the slow review cycle :)

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.lambda.events
+
+import io.circe.Decoder
+
+sealed abstract class APIGatewayV2WebSocketEvent {
```
armanbilge, line 21: We've been converting to the camel case convention in Feral.

```suggestion
sealed abstract class ApiGatewayV2WebSocketEvent {
```

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  def stageVariables: Map[String, String]
+  def requestContext: WebSocketRequestContext
+  def body: String
```
armanbilge, line 24: Looks like the body is optional, so we should use `Option[String]`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L239

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  def apiId: String
+  def connectedAt: Long
+  def connectionId: String
+  def domainName: String
+  def eventType: String
+  def extendedRequestId: String
+  def messageDirection: String
+  def messageId: String
+  def requestTime: String
+  def requestTimeEpoch: Long
```
armanbilge, line 71: I think these are redundant, two different ways of encoding the same information. We can model it as `requestTime: Instant`

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  ) extends APIGatewayV2WebSocketEvent {
+    override def productPrefix = "APIGatewayV2WebSocketEvent"
+  }
+}
+
+sealed abstract class WebSocketRequestContext {
+  def stage: String
+  def requestId: String
+  def apiId: String
+  def connectedAt: Long
```
armanbilge, line 63: If this is a time, I wonder if we should use `Instant` for it too 🤔

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  def connectionId: String
+  def domainName: String
+  def eventType: String
```
armanbilge, line 66: We can model this as an enum, since it has only three cases: `"CONNECT" | "MESSAGE" | "DISCONNECT"`

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+}
+
+sealed abstract class WebSocketRequestContext {
+  def stage: String
+  def requestId: String
+  def apiId: String
+  def connectedAt: Long
+  def connectionId: String
+  def domainName: String
```
armanbilge, line 65: Hmm, this one is interesting too. The doc says
> A domain name for the WebSocket API. This can be used to make a callback to the client (instead of a hard-coded value).

If it means "domain name" like an internet domain, we could use ip4s `Hostname` to model this. But I'm not certain if that is what it is 🤔

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
```diff
@@ -0,0 +1,140 @@
...
+  def eventType: String
+  def extendedRequestId: String
+  def messageDirection: String
```
armanbilge, line 68: This one is also an enum ... with only one value? 😂 `"IN"`. I wonder if we should just omit it, doesn't seem very useful.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L206

lambda/shared/src/main/scala/feral/lambda/events/APIGatewayV2WebSocketEvent.scala
Chingles2404: I think it should be, based on [this](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/ApiGatewayManagementApi.html)!

armanbilge (approved):
Thanks for addressing all the feedback! And nice job following the pedantic pattern we use for encoding events :)

PR #492, opened 2024-06-15
Created support for Vercel using Node.js runtime

Chingles2404:
Created the IOVercel file. Thanks Arman and Tonio for your help!

### DIFF
--- a/vercel-nodejs/src/main/scala/feral/vercel/IOVercel.scala
+++ b/vercel-nodejs/src/main/scala/feral/vercel/IOVercel.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
- package feral.vercel
+package feral.vercel
 
 import cats.effect.IO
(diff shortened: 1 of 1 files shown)


Chingles2404:
Based on discussions with Arman, it seems like Vercel is not compatible with the Http4s way of creating a web app. This is mainly due to routes being declared via the folder structure for web apps that are deployed to Vercel. Therefore, the example illustrated in `VercelNodejs.scala` is unable to work with Vercel.

The current steps to use Feral with Vercel are:
1. Write a handler for a specific route
2. Run `fastLinkJS` to generate the corresponding JS files
3. Copy the JS files into a folder named `api` which should ideally be in a separate project folder
4. Rename the JS files (and optionally put them in subfolders within `api`) to define the route
5. Deploy the project to Vercel

References:
https://vercel.com/docs/functions/runtimes/node-js
https://github.com/orgs/vercel/discussions/947

PR #496, opened 2024-07-08
feat: lambda custom runtime

i10416:
This commit resumes https://github.com/typelevel/feral/pull/276/ with some modifications.

The main differences between this commit and the PR above are the followings;

in terms of module structure,
- The original `feral-lambda` module is split into `feral-lambda-kernel` and `feral-lambda-runtime-binding`
    - `feral-lambda-kernel` contains types shared across runtime
    - `feral-lambda-runtime-binding` contains facades for AWS Java and JS runtimes
- AWS Runtime API stuffs are moved under runtime.api package
    - correcting `Lambda-Runtime-Client-Identity` to `Lambda-Runtime-Cognito-Identity` 
    - adding `LambdaRuntimeAPIClient` that wraps http4s `Client`
- Scala Native support is not included in the commit

in terms of behaviors,
- the runtime runs next invocation calls and handlers in parallel
- the runtime will terminate when Lambda Runtime API returns 500, which implies unrecoverable container error

See also
- https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html

Co-Authored-By: Scott Thomson


lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,84 @@
...
+        }
+        .foreverM[INothing]
+      val handleRequest = handleSingleRequest(client, settings, run)
+      val workers = fs2
+        .Stream
+        .fromQueueUnterminated(q)
+        .parEvalMapUnordered(1024)(handleRequest)
+        .compile
+        .drain
+      Temporal[F].both(producer, workers).map(_._1)
```
i10416, line 66: > strange assumptions such as the use of environment variables to pass around tracing headers

Running handlers in parallel conflicts with Ray SDK assumption, but I think we could ignore X-Ray specific limitation because
- if someone is happy with official Java/JS runtime and X-Ray SDK, they can use feral-lambda-runtime-binding, both of which follow the spec to propagate X-Ray header via environment variables for each invocation
- we want to support Scala Native, which AWS does not officially support at all and thus no X-Ray SDK is available for it

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,84 @@
...
+          case ex @ ContainerError => ex.raiseError[F, Unit]
+          case NonFatal(_) => ().pure
+          case ex => ex.raiseError
+        }
+        .foreverM[INothing]
+      val handleRequest = handleSingleRequest(client, settings, run)
+      val workers = fs2
+        .Stream
+        .fromQueueUnterminated(q)
+        .parEvalMapUnordered(1024)(handleRequest)
```
i10416, line 63: Should we parameterize concurrency based on ctx.lambdaMemorySize?

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
i10416: Another idea is just letting users choose qequential/concurrent execution by providing an option such as `ExecutionMode.{Parallel/Sequential}`.

armanbilge:
> the runtime runs next invocation calls and handlers in parallel

Are you sure that AWS Lambda actually support this? i.e. will the `GET` request for the next event return something different before you `POST` the result? If so, that would be violating HTTP method semantics.

lambda-kernel/shared/src/main/scala/feral/lambda/Context.scala
```diff
@@ -50,7 +50,7 @@ sealed abstract class Context[F[_]] {
 
 }
 
-object Context extends ContextCompanionPlatform {
+object Context {
```
i10416, line 53: https://github.com/awslabs/aws-lambda-rust-runtime/blob/4ee10b0906728d34926b1eba2e90702e782472ec/lambda-runtime/src/types.rs#L66

todo: some values are optional

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
i10416: According to https://github.com/RhinoSecurityLabs/Cloud-Security-Research/blob/b7d14b7eb51b7dea6ea6a8978682bd9671e0e183/AWS/lambda_ssrf/README.md and the comment in the issue, multiple call to invocation API would return the identical results 🤔 
In addition, I couldn't find any example that runs handlers in parallel...

i10416:
https://github.com/typelevel/feral/pull/496#discussion_r1669120978

I think we should avoid running handlers in parallel...
Sorry for confusion😢

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,84 @@
...
+        case Left(ex) => client.reportInitError(ex) *> ex.raiseError[F, INothing]
+      })
+
+  private def runloop[F[_]: Temporal, Event: Decoder, Result: Encoder](
+      client: LambdaRuntimeAPIClient[F],
+      settings: LambdaSettings,
+      run: Invocation[F, Event] => F[Option[Result]]) = {
+    Queue.unbounded[F, LambdaRequest].flatMap { q =>
+      val producer = client
+        .nextInvocation()
```
i10416, line 51: https://github.com/awslabs/aws-lambda-rust-runtime/blob/4ee10b0906728d34926b1eba2e90702e782472ec/lambda-runtime/src/runtime.rs#L164

should take care of 204

i10416: https://github.com/awslabs/aws-lambda-rust-runtime/blob/4ee10b0906728d34926b1eba2e90702e782472ec/lambda-runtime/src/runtime.rs#L181

should wait for the handler to be ready

lambda-kernel/shared/src/main/scala/feral/lambda/Context.scala
```diff
@@ -87,7 +119,8 @@ object Context extends ContextCompanionPlatform {
       logStreamName: String,
       identity: Option[CognitoIdentity],
       clientContext: Option[ClientContext],
-      remainingTime: F[FiniteDuration]
+      remainingTime: F[FiniteDuration],
+      xRayTraceId: Option[String]
```
i10416, line 123: propagate `X-Ray-Trace-Id` to handlers via `Context` to workaround the limitation that environment variables couldn't be set from JVM runtime.

lambda-runtime/src/main/scala/feral/lambda/runtime/LambdaRuntime.scala
```diff
@@ -0,0 +1,72 @@
...
+      settings: LambdaSettings,
+      run: Invocation[F, Event] => F[Option[Result]]) = {
+    client
+      .nextInvocation()
+      .flatMap(handleSingleRequest(client, settings, run))
+      .handleErrorWith {
+        case ex @ ContainerError => ex.raiseError[F, Unit]
+        case NonFatal(_) => ().pure
+        case ex => ex.raiseError
+      }
```
i10416, line 54: Running handlers sequentially to comply with AWS Lambda API spec and follow other implementations.

lambda-runtime/src/main/scala/feral/lambda/runtime/api/LambdaRuntimeAPIClient.scala
```diff
@@ -0,0 +1,105 @@
...
+import org.http4s.Request
+import org.http4s.Response
+import org.http4s.Status
+import org.http4s.circe.jsonEncoderOf
+import org.http4s.client.Client
+
+/**
+ * AWS Lambda Runtime API Client
+ */
+private[runtime] trait LambdaRuntimeAPIClient[F[_]] {
```
i10416, line 33: based on https://docs.aws.amazon.com/lambda/latest/dg/samples/runtime-api.zip

```
# This document describes the AWS Lambda Custom Runtime API using the OpenAPI 3.0 specification.
#
# A note on error reporting:
#
# Runtimes are free to define the format of errors that are reported to the runtime API, however,
# in order to integrate with other AWS services, runtimes must report all errors using the
# standard AWS Lambda error format:
#
# Content-Type: application/vnd.aws.lambda.error+json:
# {
#     "errorMessage": "...",
#     "errorType": "...",
#     "stackTrace": [],
# }
#
# See '#/components/schemas/ErrorRequest'.
#
# Lambda's default behavior is to use Lambda-Runtime-Function-Error-Type header value to construct an error response
# when error payload is not provided or can not be read.


openapi: 3.0.0
info:
  title: AWS Lambda Runtime API
  description: AWS Lambda Runtime API is an HTTP API for implementing custom runtimes
  version: 1.0.3

servers:
  - url: /2018-06-01

paths:

  /runtime/init/error:
    post:
      summary: >
        Non-recoverable initialization error. Runtime should exit after reporting
        the error. Error will be served in response to the first invoke.
      parameters:
        - in: header
          name: Lambda-Runtime-Function-Error-Type
          schema:
            type: string
      requestBody:
        content:
          '*/*':
            schema: {}
      responses:
        '202':
          description: Accepted
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/StatusResponse'
        '403':
          description: Forbidden
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '500':
          description: >
            Container error. Non-recoverable state. Runtime should exit promptly.

  /runtime/invocation/next:
    get:
      summary: >
        Runtime makes this HTTP request when it is ready to receive and process a
        new invoke.
      responses:
        '200':
          description: >
            This is an iterator-style blocking API call. Response contains
            event JSON document, specific to the invoking service.
          headers:
            Lambda-Runtime-Aws-Request-Id:
              description: AWS request ID associated with the request.
              schema:
                type: string
            Lambda-Runtime-Trace-Id:
              description: X-Ray tracing header.
              schema:
                type: string
            Lambda-Runtime-Client-Context:
              description: >
                Information about the client application and device when invoked
                through the AWS Mobile SDK.
              schema:
                type: string
            Lambda-Runtime-Cognito-Identity:
              description: >
                Information about the Amazon Cognito identity provider when invoked
                through the AWS Mobile SDK.
              schema:
                type: string
            Lambda-Runtime-Deadline-Ms:
              description: >
                Function execution deadline counted in milliseconds since the Unix epoch.
              schema:
                type: string
            Lambda-Runtime-Invoked-Function-Arn:
              description: >
                The ARN requested. This can be different in each invoke that
                executes the same version.
              schema:
                type: string
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/EventResponse'
        '403':
          description: Forbidden
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '500':
          description: >
            Container error. Non-recoverable state. Runtime should exit promptly.

  /runtime/invocation/{AwsRequestId}/response:
    post:
      summary: Runtime makes this request in order to submit a response.
      parameters:
        - in: path
          name: AwsRequestId
          schema:
            type: string
          required: true
      requestBody:
        content:
          '*/*':
            schema: {}
      responses:
        '202':
          description: Accepted
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/StatusResponse'
        '400':
          description: Bad Request
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '403':
          description: Forbidden
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '413':
          description: Payload Too Large
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '500':
          description: >
            Container error. Non-recoverable state. Runtime should exit promptly.

  /runtime/invocation/{AwsRequestId}/error:
    post:
      summary: >
        Runtime makes this request in order to submit an error response. It can
        be either a function error, or a runtime error. Error will be served in
        response to the invoke.
      parameters:
        - in: path
          name: AwsRequestId
          schema:
            type: string
          required: true
        - in: header
          name: Lambda-Runtime-Function-Error-Type
          schema:
            type: string
      requestBody:
        content:
          '*/*':
            schema: {}
      responses:
        '202':
          description: Accepted
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/StatusResponse'
        '400':
          description: Bad Request
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '403':
          description: Forbidden
          content:
            application/json:
              schema:
                $ref: '#/components/schemas/ErrorResponse'
        '500':
          description: >
            Container error. Non-recoverable state. Runtime should exit promptly.

components:
  schemas:

    StatusResponse:
      type: object
      properties:
        status:
          type: string

    ErrorResponse:
      type: object
      properties:
        errorMessage:
          type: string
        errorType:
          type: string

    ErrorRequest:
      type: object
      properties:
        errorMessage:
          type: string
        errorType:
          type: string
        stackTrace:
          type: array
          items:
            type: string

    EventResponse:
      type: object
```

PR #498, opened 2024-07-10
Created support for Google Cloud HTTP functions

Chingles2404:


build.sbt
```diff
@@ -224,3 +230,41 @@ lazy val scalafix = tlScalafixProject
...
+      "org.typelevel" %%% "cats-effect" % catsEffectVersion,
+      "org.tpolecat" %%% "natchez-core" % natchezVersion,
+      "io.circe" %%% "circe-scodec" % circeVersion,
+      "io.circe" %%% "circe-jawn" % circeVersion,
+      "com.comcast" %%% "ip4s-core" % "3.6.0",
+      "org.scodec" %%% "scodec-bits" % "1.2.0",
+      "org.http4s" %%% "http4s-server" % http4sVersion,
+      "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
+      "org.typelevel" %%% "munit-cats-effect-3" % munitCEVersion % Test,
+      "io.circe" %%% "circe-literal" % circeVersion % Test
```
armanbilge, line 248: Not all of these dependencies are used in our code, so we can remove some of them. For example natchez and circe.

google-cloud-http4s/jvm/src/test/scala/feral/google-cloud/http4s/IOGoogleCloudHttp.scala
```diff
@@ -0,0 +1,131 @@
...
+      def getContentLength(): Long = ???
+      def getCharacterEncoding(): ju.Optional[String] = ???
+      def getReader(): BufferedReader = ???
+      def getPath(): String = ???
+      def getQuery(): ju.Optional[String] = ???
+      def getQueryParameters(): ju.Map[String, ju.List[String]] = ???
+      def getParts(): ju.Map[String, HttpRequest.HttpPart] = ???
+    }
+
+  def googleResponse() = new HttpResponse {
```
armanbilge, line 65: I think we need to extract this to a named class:
```scala
class MockHttpResponse extends HttpResponse { /* ... */ }
```

build.sbt
```diff
@@ -224,3 +230,36 @@ lazy val scalafix = tlScalafixProject
...
+      "org.typelevel" %%% "cats-effect" % catsEffectVersion,
+      "org.scodec" %%% "scodec-bits" % "1.2.0",
+      "org.http4s" %%% "http4s-server" % http4sVersion,
+      "org.scalameta" %%% "munit-scalacheck" % munitVersion % Test,
+      "org.typelevel" %%% "munit-cats-effect-3" % munitCEVersion % Test
+    ),
+    mimaBinaryIssueFilters ++= Seq(
+      ProblemFilters.exclude[IncompatibleResultTypeProblem](
+        "feral.google-cloud.IOLambda.setupMemo")
+    ),
```
armanbilge, line 248: I think this is probably not needed?

build.sbt
```diff
@@ -224,3 +230,36 @@ lazy val scalafix = tlScalafixProject
...
+    tlVersionIntroduced := List("2.13", "3").map(_ -> "0.3.1").toMap
+  )
+  .settings(commonSettings)
+  .jsSettings(
+    libraryDependencies ++= Seq(
+      "io.circe" %%% "circe-scalajs" % circeVersion,
```
armanbilge, line 254: I don't think this dependency is used.

build.sbt
```diff
@@ -224,3 +230,36 @@ lazy val scalafix = tlScalafixProject
...
+      "io.github.cquiroz" %%% "scala-java-time" % "2.5.0"
+    )
+  )
+  .jvmSettings(
+    Test / fork := true,
+    libraryDependencies ++= Seq(
+      "com.google.cloud.functions" % "functions-framework-api" % "1.1.0",
+      "co.fs2" %%% "fs2-io" % fs2Version,
+      "com.google.cloud.functions.invoker" % "java-function-invoker" % "1.3.1"
```
armanbilge, line 263: This dependency is "provided" by Google Cloud when we deploy the function. We only need to add it when we are testing the function locally.

```suggestion
      "com.google.cloud.functions.invoker" % "java-function-invoker" % "1.3.1" % Provided
```

armanbilge (commented):
Looking good! Just a few very small things!

build.sbt
armanbilge: Turns out, this was wrong: the `functions-framework-api` should be `Provided` according to the docs, and the `invoker` should be moved to the example project because whether or not its needed at runtime depends _where_ you are running it.

build.sbt
```diff
@@ -53,11 +53,12 @@ ThisBuild / crossScalaVersions := Seq(Scala212, Scala3, Scala213)
 val catsEffectVersion = "3.5.4"
 val circeVersion = "0.14.7"
 val fs2Version = "3.10.2"
-val http4sVersion = "0.23.27"
+val http4sVersion = "0.23.27-10-fa6e976-SNAPSHOT"
```
armanbilge, line 56: We'll need to wait for http4s v0.23.28 before we can release, but we're good to merge and publish a snapshot!

armanbilge (approved):
👏

PR #552, opened 2025-03-03
Add `KafkaEvent` and `MskEvent` triggers

gfinol:
I've added the KafkaEvent. 

This is my first PR to Feral, so feedback is more than welcome (especially if there's anything incorrect).

### DIFF
--- a/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
@@ -16,7 +16,7 @@
 
 package feral.lambda.events
 
-import cats.implicits._
+import cats.syntax.all._
 import com.comcast.ip4s.Host
 import com.comcast.ip4s.SocketAddress
 import feral.lambda.events.KafkaRecord.TimestampType
--- a/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
@@ -16,6 +16,9 @@
 
 package feral.lambda.events
 
+import cats.implicits._
+import com.comcast.ip4s.Host
+import com.comcast.ip4s.SocketAddress
 import io.circe.Decoder
 import io.circe.KeyDecoder
 import scodec.bits.ByteVector
@@ -31,10 +34,12 @@ object MSKEvent {
       records: Map[TopicPartition, List[KafkaRecord]],
       eventSource: String,
       eventSourceArn: String,
-      bootstrapServers: String
-  ): MSKEvent =
+      bootstrapServers: List[SocketAddress[Host]]
+  ): MSKEvent = {
     Impl(records, eventSource, eventSourceArn, bootstrapServers)
+  }
 
+  import feral.lambda.events.KafkaEvent.bootstrapServersDecoder
   private[events] implicit val decoder: Decoder[MSKEvent] =
     Decoder.forProduct4(
       "records",
@@ -47,7 +52,7 @@ object MSKEvent {
       records: Map[TopicPartition, List[KafkaRecord]],
       eventSource: String,
       eventSourceArn: String,
-      bootstrapServers: String
+      bootstrapServers: List[SocketAddress[Host]]
   ) extends MSKEvent {
     override def productPrefix = "KafkaEvent"
   }
@@ -56,16 +61,28 @@ object MSKEvent {
 sealed abstract class KafkaEvent {
   def records: Map[TopicPartition, List[KafkaRecord]]
   def eventSource: String
-  def bootstrapServers: String
+  def bootstrapServers: List[SocketAddress[Host]]
 }
 
 object KafkaEvent {
   def apply(
       records: Map[TopicPartition, List[KafkaRecord]],
       eventSource: String,
-      bootstrapServers: String
-  ): KafkaEvent =
+      bootstrapServers: List[SocketAddress[Host]]
+  ): KafkaEvent = {
     Impl(records, eventSource, bootstrapServers)
+  }
+
+  private[events] implicit val bootstrapServersDecoder: Decoder[List[SocketAddress[Host]]] =
+    Decoder
+      .decodeString
+      .emap(str =>
+        str
+          .split(",")
+          .toList
+          .map(SocketAddress.fromString)
+          .sequence
+          .toRight(s"Failed to parse bootstrap servers: $str"))
 
   private[events] implicit val decoder: Decoder[KafkaEvent] =
     Decoder.forProduct3(
@@ -77,7 +94,7 @@ object KafkaEvent {
   private final case class Impl(
       records: Map[TopicPartition, List[KafkaRecord]],
       eventSource: String,
-      bootstrapServers: String
+      bootstrapServers: List[SocketAddress[Host]]
   ) extends KafkaEvent {
     override def productPrefix = "KafkaEvent"
   }
@@ -88,7 +105,7 @@ sealed abstract class KafkaRecord {
   def partition: Int
   def offset: Long
   def timestamp: Instant
-  def timestampType: String
+  def timestampType: TimestampType
   def key: ByteVector
   def value: ByteVector
   def headers: List[Map[String, ByteVector]]
@@ -100,7 +117,7 @@ object KafkaRecord {
       partition: Int,
       offset: Long,
       timestamp: Instant,
-      timestampType: String,
+      timestampType: TimestampType,
       key: ByteVector,
       value: ByteVector,
       headers: List[Map[String, List[Int]]]
@@ -130,7 +147,7 @@ object KafkaRecord {
       partition: Int,
       offset: Long,
       timestamp: Instant,
-      timestampType: String,
+      timestampType: TimestampType,
       key: ByteVector,
       value: ByteVector,
       headers: List[Map[String, ByteVector]]
@@ -165,5 +182,17 @@ object TopicPartition {
       partition: Int
   ) extends TopicPartition {
     override def productPrefix = "TopicPartition"
+  }
+}
+
+sealed trait TimestampType
+case object CREATE_TIME extends TimestampType
+case object LOG_APPEND_TIME extends TimestampType
+
+object TimestampType {
+  implicit val decoder: Decoder[TimestampType] = Decoder.decodeString.emap {
+    case "CREATE_TIME" => Right(CREATE_TIME)
+    case "LOG_APPEND_TIME" => Right(LOG_APPEND_TIME)
+    case other => Left(s"Unknown timestamp type: $other")
   }
 }
--- a/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
@@ -30,6 +30,9 @@ sealed abstract class MSKEvent extends KafkaEvent {
 }
 
 object MSKEvent {
+  import KafkaEvent.TopicPartition
+  import KafkaEvent.bootstrapServersDecoder
+
   def apply(
       records: Map[TopicPartition, List[KafkaRecord]],
       eventSourceArn: String,
@@ -38,7 +41,6 @@ object MSKEvent {
     Impl(records, eventSourceArn, bootstrapServers)
   }
 
-  import feral.lambda.events.KafkaEvent.bootstrapServersDecoder
   private[events] implicit val decoder: Decoder[MSKEvent] =
     Decoder.forProduct3(
       "records",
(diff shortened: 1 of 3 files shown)


armanbilge:
> There are some fields that appear in one example but not in the other.

I think that's because they are actually examples for two different event types :)

1. Amazon MSK
  https://docs.aws.amazon.com/lambda/latest/dg/with-msk.html#msk-sample-event
2. self-managed Apache Kafka
  https://docs.aws.amazon.com/lambda/latest/dg/with-kafka.html#smaa-sample-event

> I couldn't find an official "event specification" listing all field descriptions.

The TypeScript definitions are another good source of "event specification".

https://github.com/DefinitelyTyped/DefinitelyTyped/tree/b11fef11f3f6e3a45e25a02f03a9fc9f7e204393/types/aws-lambda/trigger

gfinol:
Ups... I missed that! 

Using the TypeScript definitions, I created another type for MSK events, which adds the `eventSourceArn` to the `KafkaEvent`. 

It is ready for review, let me know what do you think

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,169 @@
...
+    override def productPrefix = "KafkaEvent"
+  }
+}
+
+sealed abstract class KafkaRecord {
+  def topic: String
+  def partition: Int
+  def offset: Long
+  def timestamp: Instant
+  def timestampType: String
```
armanbilge, line 91: I guess this only has two possible values, so we can model it as an ADT.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b11fef11f3f6e3a45e25a02f03a9fc9f7e204393/types/aws-lambda/trigger/msk.d.ts#L14

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,169 @@
...
+      eventSourceArn: String,
+      bootstrapServers: String
+  ) extends MSKEvent {
+    override def productPrefix = "KafkaEvent"
+  }
+}
+
+sealed abstract class KafkaEvent {
+  def records: Map[TopicPartition, List[KafkaRecord]]
+  def eventSource: String
```
armanbilge, line 58: Similarly, it seems that `eventSource` has known values for each of the event types. Honestly, I'm not sure how much value this field adds.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b11fef11f3f6e3a45e25a02f03a9fc9f7e204393/types/aws-lambda/trigger/msk.d.ts#L14

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b11fef11f3f6e3a45e25a02f03a9fc9f7e204393/types/aws-lambda/trigger/self-managed-kafka.d.ts#L22

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,169 @@
...
+  def bootstrapServers: String
```
armanbilge, line 59: Based on the examples, this looks like it should be a `List[SocketAddress[Host]]`
https://github.com/Comcast/ip4s/blob/0ca38a4e40d7f92f67460eba1f709e28971fcb55/shared/src/main/scala/com/comcast/ip4s/SocketAddress.scala#L24

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
gfinol: True, I'll do it!

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
gfinol: Totally!

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
gfinol: Yes... I am not sure either... 

I think we have three options here:
1. Parse the value
2. Assume it will be always what it is expected
3. Delete the field

Not sure which one is the best, WDYT?

armanbilge: I'm okay with (3), we can always add it in the future if it turns out to be relevant somehow.

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,191 @@
...
+      timestampType: TimestampType,
+      key: ByteVector,
+      value: ByteVector,
+      headers: List[Map[String, ByteVector]]
+  ) extends KafkaRecord {
+    override def productPrefix = "KafkaEventRecord"
+  }
+}
+
+sealed abstract class TopicPartition {
```
armanbilge, line 152: Similarly, we can nest this under `KafkaEvent`.

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,191 @@
...
+
+sealed abstract class KafkaRecord {
+  def topic: String
+  def partition: Int
+  def offset: Long
+  def timestamp: Instant
+  def timestampType: TimestampType
+  def key: ByteVector
+  def value: ByteVector
+  def headers: List[Map[String, ByteVector]]
```
armanbilge, line 104: This `List[Map[...]]` feels really strange to me ... I don't understand why it's not just `Map[`. Sorry, I should say I'm not familiar with Kafka at all.

Maybe they use this weird representation to try and preserve order? If so, we have better ways to represent that in Scala.

gfinol: I think it is modeled as a `List[Map[...]]` because [Kafka Headers may contain duplicates keys](https://kafka.apache.org/20/javadoc/org/apache/kafka/connect/header/Header.html#key--) (that's a bad practice, but Kafka allows it).

gfinol: I think so... 

There are other options that could also make sense, like `Map[String, List[ByteVector]]`, or just assuming that there are no duplicates with `Map[]` ... But maybe going for the simpler one (`List[(String, ByteVector)]`) is the best option here...

armanbilge: Unfortunately, probably none of those options quite work 😕 we have to allow duplicate headers and we also have to preserve the order.

> Key duplication and ordering
> - Duplicate headers with the same key must be supported.
> - The order of headers must be retained throughout a record's end-to-end lifetime: from producer to consumer.

https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=65871723#KIP82AddRecordHeaders-Keyduplicationandordering

gfinol: The Kafka API for Java uses [Headers](https://github.com/a0x8o/kafka/blob/master/connect/api/src/main/java/org/apache/kafka/connect/header/Headers.java) that is just an iterable of Header. 

I think that given the JSON schema that AWS provides, we should assume that AWS Lambda will give us the Headers ordered inside the Map, with duplicate headers in the following Maps... 

It is the best that we can do, isn't?

armanbilge: > The Kafka API for Java uses [Headers](https://github.com/a0x8o/kafka/blob/master/connect/api/src/main/java/org/apache/kafka/connect/header/Headers.java) that is just an iterable of Header.

Aha, then yes, a `List[(String, ByteVector)]` is appropriate.

> I think that given the JSON schema that AWS provides, we should assume that AWS Lambda will give us the Headers ordered inside the Map, with duplicate headers in the following Maps...

Oh, I was assuming that each map will only contain one pair ... essentially, it's just a way to represent the key/value pair. Especially because there isn't an ordering guarantee for keys.

gfinol: > I was assuming that each map will only contain one pair
Yep... That would make sense. In that case, I think that the current implementation should work as expected.

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,206 @@
...
+      key: ByteVector,
+      value: ByteVector,
+      headers: List[(String, ByteVector)]
+  ) extends KafkaRecord {
+    override def productPrefix = "KafkaEventRecord"
+  }
+
+  sealed trait TimestampType
+  case object CREATE_TIME extends TimestampType
+  case object LOG_APPEND_TIME extends TimestampType
```
armanbilge, line 197:
```suggestion
  sealed abstract class TimestampType
  object TimestampType {
    case object CreateTime extends TimestampType
    case object LogAppendTime extends TimestampType
  }
```

armanbilge (commented):
Looking good! Just some final adjustments.

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,207 @@
...
+      value: ByteVector,
+      headers: List[(String, ByteVector)]
+  ): KafkaRecord = {
+    Impl(topic, partition, offset, timestamp, timestampType, key, value, headers)
+  }
+
+  private[events] implicit val headersDecoder: Decoder[List[(String, ByteVector)]] = {
+    val byteHeadersDecoder: Decoder[ByteVector] = Decoder.decodeList(Decoder.decodeInt).map {
+      list => ByteVector(list.map(_.toByte).toArray)
+    }
```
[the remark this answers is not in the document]
gfinol: The compiler complains about "Unspecified value parameters: factory: `Factory[NotInferredA, Array[NotInferredA]]`"

So I passed explicitly `Array`, what do you think?
```
val byteHeadersDecoder: Decoder[ByteVector] = Decoder.decodeArray(Decoder.decodeByte, Array).map(ByteVector(_))
```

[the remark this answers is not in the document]
armanbilge: Oh, hmm ... maybe let's do `decodeArray(Decoder.decodeByte, ClassTag.Byte)` ? I have more confidence that this will avoid boxing.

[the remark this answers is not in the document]
gfinol: mmm... The compiler does not it...

Type mismatch. Required: `Factory[NotInferredA, Array[NotInferredA]]`, found: `ManifestFactory.ByteManifest`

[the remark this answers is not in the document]
armanbilge: Gah, sorry! I was look at the wrong API, I was looking at the one for [`ArraySeq`](https://github.com/circe/circe/blob/f5407111aa376e7d71f723a43adac67de8f97c1f/modules/core/shared/src/main/scala-2.13%2B/io/circe/CollectionDecoders.scala#L80). My fault :)

> So I passed explicitly `Array`, what do you think?

Yes, that's right. It looks like it also uses `ClassTag` under-the-hood.

https://github.com/scala/scala3/blob/ad3f69867b4600a3550bf9e76a1fc58091cb5eaf/scala2-library-bootstrapped/src/scala/Array.scala#L47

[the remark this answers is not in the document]
gfinol: Done!

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,205 @@
...
+      topic: String,
+      partition: Int,
+      offset: Long,
+      timestamp: Instant,
+      timestampType: TimestampType,
+      key: ByteVector,
+      value: ByteVector,
+      headers: List[(String, ByteVector)]
+  ) extends KafkaRecord {
+    override def productPrefix = "KafkaEventRecord"
```
armanbilge, line 191:
```suggestion
    override def productPrefix = "KafkaRecord"
```

lambda/shared/src/main/scala/feral/lambda/events/KafkaEvent.scala
```diff
@@ -0,0 +1,205 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package feral.lambda.events
+
+import cats.implicits._
```
armanbilge, line 19:
```suggestion
import cats.syntax.all._
```

armanbilge (approved):
I think this is ready to go! Good job, these events are not the best documented, so it takes a lot of investigation sometimes 😅

PR #639, opened 2026-03-12
feature: Added APIGatewayAuthorizerEvent & results

Yummy-Yums:
#48  : adds AWS Lambda event type APIGatewayAuthorizerEvent:

image


Yummy-Yums:
@armanbilge please review

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
```diff
@@ -0,0 +1,208 @@
...
+      domainName: Hostname,
+      deploymentId: String,
+      apiId: String
+  ) extends AuthorizerRequestContext {
+    override def productPrefix = "AuthorizerRequestContext"
+  }
+}
+
+sealed abstract class ApiGatewayCustomAuthorizerEvent {
+  def `type`: String
```
Chingles2404: Hi Elias, great job! 

I'm wondering if it would be better to have this as a custom object instead of String, similar to [here](https://github.com/typelevel/feral/blob/017f4aebbe53c3235071d8e45ffb04bd12f6bdad/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayV2WebSocketEvent.scala#L67), since "type" can only be "TOKEN" and "REQUEST" ([reference](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-input.html)).

I also noticed that DefinitelyTyped uses a single [RequestContext](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L45) function for both the proxy events and the authorizer ones, so I am also wondering if, in the future, unifying the implementations would be better than having separate ones (like the one in [WebSocketEvent](https://github.com/typelevel/feral/blob/017f4aebbe53c3235071d8e45ffb04bd12f6bdad/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayV2WebSocketEvent.scala#L80)).

Chingles2404 (changes requested):

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
Chingles2404: > I also noticed that DefinitelyTyped uses a single [RequestContext](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L45) function for both the proxy events and the authorizer ones, so I am also wondering if, in the future, unifying the implementations would be better than having separate ones (like the one in [WebSocketEvent](https://github.com/typelevel/feral/blob/017f4aebbe53c3235071d8e45ffb04bd12f6bdad/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayV2WebSocketEvent.scala#L80)).

Oh this was meant for @armanbilge, sorry I didn't mention you!

Chingles2404 (approved):
Thanks!

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
```diff
@@ -0,0 +1,221 @@
...
+
+import codecs.decodeHostname
+import codecs.decodeKeyCIString
+
+sealed abstract class AuthorizerRequestContext {
+  def resourceId: String
+  def resourcePath: String
+  def httpMethod: String
+  def extendedRequestId: String
+  def requestTime: String
```
armanbilge, line 34: I see there is `requestTime: String` and `requestTimeEpoch: Long`. Probably they are redundant, and we could use `requestTime: Instant` for the best UX?

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
```diff
@@ -0,0 +1,221 @@
...
+  def path: String
+  def accountId: String
+  def protocol: String
+  def stage: String
+  def domainPrefix: String
+  def requestTimeEpoch: Long
+  def requestId: String
+  def identity: Map[String, Option[String]]
```
armanbilge, line 42: This type looks a bit strange 🤔 I wonder in what case would there be a key would no value? (instead of just not having the key at all).

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
armanbilge: > I also noticed that DefinitelyTyped uses a single [RequestContext](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L45) function for both the proxy events and the authorizer ones

Great catch! I agree, instead of defining a new class here, maybe let's re-use the existing `RequestContext` class and add all the additional fields. I'm not sure why it is currently missing most of them 🤔 

https://github.com/typelevel/feral/blob/017f4aebbe53c3235071d8e45ffb04bd12f6bdad/lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayProxyEventV2.scala#L40-L42

armanbilge (changes requested):
Great idea from Ching about sharing the `RequestContext` class, let's do that!

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
Yummy-Yums: From the [Input format of a Lambda function for proxy integration](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format) there are such cases . Also, I'd specifically update the `identity` key  to use an identity object as done here 

Image

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
Yummy-Yums: I don't think they are redundant since it's enumerated  here in the output format

Image

armanbilge: What I mean by "redundant" is that the field contain identical information, just presented in different formats. We should be able to retrieve the time entirely using the epoch timestamp.

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
armanbilge: Interesting, I see. It looks like it would be better to define a new class for `identity` instead of using a `Map` since the fields appear to be predefined.

lambda/shared/src/main/scala/feral/lambda/events/ApiGatewayCustomAuthorizerEvent.scala
Chingles2404: ### AWS Lambda Documentation
Based on what I could find from the Lambda docs, there are two RequestContext types labelled under data transformations, one for [REST APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference) and the other for [WebSocket APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-websocket-api-mapping-template-reference.html). There are also three others labelled under access logging, for [REST APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-variables-for-access-logging.html), [WebSocket APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/websocket-api-logging.html), and [HTTP APIs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-logging-variables.html).

For the [Authorizer event V2](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-lambda-authorizer.html) (note that this PR is for V1), the RequestContext also has a HTTP portion, but I have been unable to find any documentation that contains `$context.http.[variables]`. It seems like the both [TypeScript](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L174) and [Java](https://github.com/aws/aws-lambda-java-libs/blob/cfefd3bb692597d1f5effa9c99f84021cd35f0e0/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2CustomAuthorizerEvent.java#L16) implementations just reference this page though, so I'm guessing that there probably aren't any additional variables for this RequestContext listed elsewhere.

### Implementations
Looking at V1 and V2 for the gateway events...

The Java implementation seems to have a RequestContext class for practically every API Gateway event.
1. [APIGatewayCustomAuthorizerEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayCustomAuthorizerEvent.java)
2. [APIGatewayV2CustomAuthorizerEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2CustomAuthorizerEvent.java)
3. [APIGatewayProxyRequestEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayProxyRequestEvent.java)
4. [APIGatewayV2HTTPEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2HTTPEvent.java)
5. [APIGatewayV2WebSocketEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2WebSocketEvent.java) ([APIGatewayV2ProxyRequestEvent](https://github.com/aws/aws-lambda-java-libs/blob/main/aws-lambda-java-events/src/main/java/com/amazonaws/services/lambda/runtime/events/APIGatewayV2ProxyRequestEvent.java) inherits this class i.e. inherits its RequestContext class too but it's apparently because it's deprecated)

The TypeScript implementation seems to have three main RequestContext classes.
1. [APIGatewayEventRequestContextWithAuthorizer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/common/api-gateway.d.ts#L26) (used by [APIGatewayRequestAuthorizerEvent](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L91) and [APIGatewayProxyEventBase](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L146)) (this seems to follow the $context variables outlined in the data transformations for REST APIs)
2. [APIGatewayEventRequestContextV2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L174) (used by [APIGatewayRequestAuthorizerEventV2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-authorizer.d.ts#L45), [APIGatewayProxyEventV2WithJWTAuthorizer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L278), [APIGatewayProxyEventV2WithLambdaAuthorizer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L282), [APIGatewayProxyEventV2WithIAMAuthorizer](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L289), [APIGatewayProxyEventV2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L300)) (this seems to follow the RequestContext on the page for Authorizer event V2)
3. [APIGatewayEventWebsocketRequestContextV2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L200) (used by [APIGatewayProxyWebsocketEventV2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e40cab53d9971a3e0b34e8fb18ece74240d9425d/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L305)) (this seems to follow the $context variables outlined in the data transformations for WebSocket APIs)

If this information is correct, how should we proceed? @armanbilge

PR #642, opened 2026-03-16
Add `AppSyncLambdaAuthorizer`

Yummy-Yums:
https://github.com/typelevel/feral/issues/48 : adds AWS Lambda event type AppSynclambdaAuthorizer

### DIFF
--- a/lambda/shared/src/main/scala/feral/lambda/events/AppSyncLambdaAuthorizerEvent.scala
+++ b/lambda/shared/src/main/scala/feral/lambda/events/AppSyncLambdaAuthorizerEvent.scala
@@ -20,86 +20,83 @@ package events
 import io.circe.Decoder
 
 sealed abstract class AppSyncRequestContext {
-    def apiId: String
-    def accountId: String
-    def requestId: String
-    def operation: String
-    def channelNamespaceName: String
-    def channel: String
+  def apiId: String
+  def accountId: String
+  def requestId: String
+  def operation: String
+  def channelNamespaceName: String
+  def channel: String
 }
 
 object AppSyncRequestContext {
 
-    def apply(
-        apiId: String,
-        accountId: String,
-        requestId: String,
-        operation: String,
-        channelNamespaceName: String,
-        channel: String
+  def apply(
+      apiId: String,
+      accountId: String,
+      requestId: String,
+      operation: String,
+      channelNamespaceName: String,
+      channel: String
+  ): AppSyncRequestContext =
+    new Impl(
+      apiId,
+      accountId,
+      requestId,
+      operation,
+      channelNamespaceName,
+      channel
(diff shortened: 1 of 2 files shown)


Yummy-Yums:
@armanbilge pleas review

Chingles2404 (approved):
Looks great! Thanks for your hard work!

lambda/shared/src/main/scala/feral/lambda/events/AppSyncLambdaAuthorizerEvent.scala
```diff
@@ -0,0 +1,103 @@
...
+      variables: Map[String, Json]
+  ) extends AppSyncRequestContext {
+    override def productPrefix = "AppSyncRequestContext"
+  }
+}
+
+sealed abstract class AppSyncLambdaAuthorizerEvent {
+  def authorizationToken: String
+  def requestContext: AppSyncRequestContext
+  def requestHeaders: Map[String, String]
```
armanbilge: If these are HTTP headers, we can use `CIString` because they are case-insensitive.

```suggestion
  def requestHeaders: Map[CIString, String]
```

armanbilge (changes requested):
Looks good, noticed one thing!

lambda/shared/src/test/scala/feral/lambda/events/AppSyncLambdaAuthorizerEventSuite.scala
```diff
@@ -0,0 +1,56 @@
...
+  test("decoder") {
+    import AppSyncLambdaAuthorizerEventSuite._
+
+    val parsed_event = event.as[AppSyncLambdaAuthorizerEvent].toTry.get
+
+    assertEquals(parsed_event.authorizationToken, "ExampleAUTHtoken123123123")
+    assertEquals(parsed_event.requestContext.accountId, "111122223333")
+    assertEquals(parsed_event.requestContext.apiId, "aaaaaa123123123example123")
+    assertEquals(parsed_event.requestContext.requestId, "f4081827-1111-4444-5555-5cf4695f339f")
+    assertEquals(parsed_event.requestHeaders(CIString("header")), "value")
```
armanbilge: Sorry, one last comment (it applies to all of your PRs): instead of checking each field individually, could we construct an instance of the object and assert its equality like in the existing tests? I think this is a preferable approach because we can be confident we didn't forget to test any fields, etc.

https://github.com/typelevel/feral/blob/017f4aebbe53c3235071d8e45ffb04bd12f6bdad/lambda/shared/src/test/scala/feral/lambda/events/S3EventSuite.scala#L28

armanbilge (changes requested):

armanbilge (approved):

PR #658, opened 2026-06-12
Add Google Cloud Events Pubsub & Storage

Yummy-Yums:
Google Cloud Run Platform Events for [Storage](https://github.com/googleapis/google-cloudevents/blob/main/jsonschema/google/events/cloud/storage/v1/StorageObjectData.json) & [Pubsub](https://github.com/googleapis/google-cloudevents/blob/main/jsonschema/google/events/cloud/pubsub/v1/MessagePublishedData.json)

### DIFF
--- a/google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
+++ b/google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package feral.`google-cloud`
+package feral.googlecloud
 package events
 
 import io.circe._
--- a/google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
+++ b/google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
@@ -19,6 +19,8 @@ package events
 
 import io.circe._
 
+import java.time.Instant
+
 sealed abstract class MessagePublishedData {
   def message: PubsubMessage
   def subscription: String
@@ -53,7 +55,7 @@ sealed abstract class PubsubMessage {
   def data: String
   def attributes: Map[String, String]
   def messageId: String
-  def publishTime: String
+  def publishTime: Instant
   def orderingKey: Option[String]
 }
 
@@ -63,11 +65,13 @@ object PubsubMessage {
       data: String,
       attributes: Map[String, String],
       messageId: String,
-      publishTime: String,
+      publishTime: Instant,
       orderingKey: Option[String]
   ): PubsubMessage = {
     new Impl(data, attributes, messageId, publishTime, orderingKey)
   }
+
+  import codecs.decodeInstant
 
   implicit val decoder: Decoder[PubsubMessage] = Decoder.forProduct5(
     "data",
@@ -81,7 +85,7 @@ object PubsubMessage {
       data: String,
       attributes: Map[String, String],
       messageId: String,
-      publishTime: String,
+      publishTime: Instant,
       orderingKey: Option[String]
   ) extends PubsubMessage { override def productPrefix = "PubsubMessage" }
 }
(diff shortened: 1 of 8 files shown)


google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
```diff
@@ -0,0 +1,87 @@
...
+      message: PubsubMessage,
+      subscription: String,
+      deliveryAttempt: Option[Int]
+  ) extends MessagePublishedData {
+    override def productPrefix = "MessagePublishedData"
+  }
+}
+
+sealed abstract class PubsubMessage {
+  def data: String
```
Chingles2404: I think this could probably be a ByteVector? ([Code ref](https://github.com/googleapis/google-cloudevents/blob/6fda292a458d2d1a534e6faaf8bd766cf9e32eb6/proto/google/events/cloud/pubsub/v1/data.proto#L51))

(From the [JSON schema](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/pubsub/v1/MessagePublishedData.json))
```json
"data": {
          "type": "string",
          "description": "The binary data in the message.",
          "format": "binary",
          "binaryEncoding": "base64"
        }

File: google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala [hunk 2]
```diff
@@ -0,0 +1,87 @@
...
+  def attributes: Map[String, String]
+  def messageId: String
+  def publishTime: String
```
Chingles2404: Would it be better to have a custom object for this like [Google's](https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto)? ([Usage](https://github.com/googleapis/google-cloudevents/blob/6fda292a458d2d1a534e6faaf8bd766cf9e32eb6/proto/google/events/cloud/pubsub/v1/data.proto#L62))

Chingles2404 (changes requested):

google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
Yummy-Yums: umm..the code ref linked is a protobuf-generated ..and it's a string of bytes .

google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
Yummy-Yums: that's also protobuf generated. I'd use Instant just as @armanbilge suggested [here](https://github.com/typelevel/feral/pull/639#discussion_r3482478916)

google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
armanbilge: I agree with Ching, I believe this should be a `ByteVector` that is parsed from its base64 representation.

Chingles2404 (changes requested):
For `StorageObjectData.scala`, could you go through the fields and change the fields with `"format": "date-time"` in the [JSON definition](https://googleapis.github.io/google-cloudevents/jsonschema/google/events/cloud/storage/v1/StorageObjectData.json) to Instant too?

Other than this, I think it looks great!

Chingles2404 (approved):
Looks good to me! Thanks for your work!

Yummy-Yums:
@armanbilge

google-cloud/shared/src/main/scala/feral/google-cloud/events/MessagePublishedData.scala
```diff
@@ -0,0 +1,92 @@
...
+  private final case class Impl(
+      message: PubsubMessage,
+      subscription: String,
+      deliveryAttempt: Option[Int]
+  ) extends MessagePublishedData {
+    override def productPrefix = "MessagePublishedData"
+  }
+}
+
+sealed abstract class PubsubMessage {
```
armanbilge:
```suggestion
sealed abstract class PubSubMessage {
```

